PR #10, opened 2021-03-13
Add bulma adder

manuel3108:
As discussed on discord, here is the bulma adder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ They are all *composable*, meaning that it should always be possible to run one 
 
 * [**svelte-add/tailwindcss**](https://github.com/svelte-add/tailwindcss): Add Tailwind CSS to your SvelteKit project
 
+* [**manuel3108/svelte-add-bulma**](https://github.com/manuel3108/svelte-add-bulma): Add Bulma to your SvelteKit project
+
 ## 💡 Example
 To migrate from [`sapper-firebase-typescript-graphql-tailwindcss-actions-template`](https://github.com/babichjacob/sapper-firebase-typescript-graphql-tailwindcss-actions-template) to SvelteKit, these commands can be run to recreate all the functionality:
 


babichjacob:
Here are my thoughts:
* Change "Finally, run this command in your project directory to set up Tailwind CSS:" to "Once that is set up, run this command in your project directory to set up Bulma:"
* Change the 💨 emoji to something that better represents Bulma, like 📗 or 🍃 or some other green emoji (or one that isn't the same color as their logo if it better evokes the feeling of Bulma, so to speak)
* Change the `$layout.svelte` file from
  ```svelte
  <script>
      let temp = "";
  </script>
  
  <slot />
  
  <style global lang="scss">
      @import "../bulma.scss";
  </style>
  ```
  
  to
  ```html
  <script>
      import "../bulma.scss";
  </script>
  
  <slot />
  ```
  This prevents the Svelte compiler from being overloaded with a bunch of CSS.

* I'm of the opinion that demos should always be included in just 1-2 files instead of asking the user for approval. The `Buttons.svelte`, `Form.svelte`, and `Table.svelte` can all be put into `bulma-demo.svelte` directly. Or a large variety of components (instead of just those 3) can be put in `bulma-demo.svelte` instead of showing variations on the same ones (like is currently happening in `Button.svelte`).
* `bulma.scss` may want to be renamed `global.scss` so it matches `svelte-add/postcss`'s `global.css`. 
* I think [the instructions at the end of the preset](https://github.com/manuel3108/svelte-add-bulma/blob/main/preset.ts#L61-L67) are better put in the README only since people can read the README without having to apply the preset. 

We can transfer this to the `svelte-add` organization—if you want—when this is all addressed.

manuel3108:
Thanks for the Feedback, I agree with you in most of the points.
I have already updated my repo.

I will open a few issues shortly to address the things that i have found. They might be interesting for more people

babichjacob:
Looks good to me! You can transfer it to `svelte-add/bulma` and be added as a collaborator if you'd like.

manuel3108:
Yeah, I think that would be great :+1:

babichjacob:
> Yeah, I think that would be great 👍

I don't seem to have gotten a notification anywhere of a transfer attempt. Is there something special [this GitHub support page](https://docs.github.com/en/github/administering-a-repository/transferring-a-repository#transferring-a-repository-owned-by-your-user-account) says there is to do?

manuel3108:
Didn't tried jet because I thought I would need some permissions.

I just tried it nevertheless, but it doesn't work:


> To transfer repositories to an organization, you must have repository creation permissions in the receiving organization. If organization owners have disabled repository creation by organization members, only organization owners can transfer repositories out of or into the organization.

From my understanding you would need to give me some permissions. Otherwise I think you could create the repo by yourself and just add me there as a contributor.
Na other way would probably be that I give you access to my repo, so that you can transfer it

babichjacob:
I think if you transfer it to me I can transfer it to the organization. Also, I am sorry that my schoolwork is going to be really important for the next couple days so things like this are going to be deprioritized.

manuel3108:
Yea, sure no problem.
I have requested the transfer, but take the time you need

babichjacob:
> Yea, sure no problem.
> I have requested the transfer, but take the time you need

It's up, links are updated, and you've been invited as an admin. 🥳

PR #19, opened 2021-04-06
Add CoffeeScript adder.

Leftium:
I made a CoffeeScript adder. How do I finish contributing it?

Tested with SvelteKit. I'm not sure if non-SvelteKit Vite works; I tried to leave the necessary code in there.

Special highlight: this adder makes an effort to preserve existing sveltePreprocess() options:

https://github.com/Leftium/coffeescript-adder/blob/cc0f2004cbc00c3224b0c095d01bd3a068f0eaef/preset.ts#L3-L14

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ They are all *composable*, meaning that it should always be possible to run one 
 
 * [**svelte-add/postcss**](https://github.com/svelte-add/postcss): Add PostCSS to your SvelteKit project
 
+* [**leftium/pug-adder**](https://github.com/Leftium/pug-adder): Add Pug to your SvelteKit project
+
 * [**svelte-add/tailwindcss**](https://github.com/svelte-add/tailwindcss): Add Tailwind CSS to your SvelteKit project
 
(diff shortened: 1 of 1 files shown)


babichjacob:
Hey! I'd be happy to help you out. It's just that, right now, school is quite overwhelming.

I'm a bit paranoid about the `eval` call so I think we'll leave this out of the `svelte-add` organization until I can create the solution mentioned in https://github.com/svelte-add/postcss/issues/17.

By the way, if you want people to use the adder as it is now (outside of the `svelte-add` organization), the command is actually
```sh
npx svelte-add Leftium/coffeescript-adder
```

Leftium:
OK, I updated my repo to use `Leftium/coffeescript-adder` until it's added to the org.

I think `eval` is very low risk because it's only executed when running `svelte-add`. To attack, another adder would have to put malicious code inside svelte.config.cjs. It would be simpler to just attack from the malicious adder itself.

Looking forward to your solution!

babichjacob:
Thanks! I'm merging this in its current state even though the problems aren't resolved. That's a future thing for me to do.

Issue #17, opened 2021-03-17
project initialization with svelte adders


manuel3108:
This afternoon I had a though: How would it be, if we could immediately add some of those awesome adders while running `npm init svelte@next`

I found that this is quite an interesting question as this would help people setting up stuff really quickly. So there we are:
```
npm init svelte-extended
```
This command initializes a new sveltekit project, reads all available repos of this org, and gives the users the possibility to immediately install them. For who does not have a terminal at his hands, here is an image:
![image](https://user-images.githubusercontent.com/30698007/111529036-011aa400-8762-11eb-9945-90bba64fc524.png)

The repo for this stuff currently is here: https://github.com/manuel3108/create-svelte-extended (for anyone who wants to have a look). It's currently not tuned and probably has many problems. (Under the hood, this all just executes npm / npx commands)

What are your thoughts about that? Any ideas, problems etc?

PS: I'm writing this here, because it uses all the templates of this org. Depending on your thoughts and the problems you see, we could maybe also transfer this to this org.

babichjacob:
I will absolutely explore this in the future!

What I'm expecting it to look like is
```sh
npm init @svelte-add/kit  # Create a SvelteKit project (their prompts) then launch the Svelte Adders selection menu 
# or
npm init @svelte-add/vite  # Create a Vite-powered Svelte app then launch the Svelte Adders selection menu 
```

babichjacob:
Related: https://github.com/svelte-add/svelte-adders/issues/12

babichjacob:
Closed by 09c11d5c5f9117eeca8fc5120943439a3b3e30c3. There may be some problems, so I'm leaving this issue open for a while.

manuel3108:
@babichjacob I think you either referenced the wrong commute or wrote in the wrong issue, or I don't understand the commit ;)

babichjacob:
> @babichjacob I think you either referenced the wrong commute or wrote in the wrong issue, or I don't understand the commit ;)


You're right, this is a better commit to look at:

https://github.com/svelte-add/svelte-add/commit/34876aaed8f90463c6c8475179719857e8e7adda#diff-956c9fec778dee35effe338392c21cd77daa31ca3391ae168d6d7d29efd6a0bdR252-R485

babichjacob:
> Hey, just got some time to test it out. First of all, very well done!!! The process is smooth, the questions and selections are very clear and overall it "just works".

Thank you!!! I really appreciate you taking the time to check it out.


[...]

PR #140, opened 2021-09-18
Rework bulma adder

manuel3108:
See https://github.com/svelte-add/bulma/pull/8, #17 

As soon as the other one is merged, we will need update submodule version to take the commit from the main branch.

Honestly, I don't even know what happens if you check that out, will it take the files from the https://github.com/svelte-add/bulma/tree/rework-bulma-adder branch? If not, you may need to change to that branch accordingly


manuel3108:
Well, thats what I guessed, pipeline is propably not happy, because the commit im referencing in .gitmodules is not on the default branch. No idea on how to fix that

babichjacob (approved):

babichjacob:
I'll work it out.

PR #154, opened 2021-10-02
add github pages

manuel3108:
Github pages can be used to host your static svelte-kit apps.
For more details see the readme [here](https://github.com/manuel3108/svelte-add-github-pages/blob/main/README.md).

I created a new method in the main `svelte-add` package to add an adapter to the current svelte-kit installation. We might need to add some checks, to see if the current application is really a `kit` application.

Other changes are minor.
Pipeline should work and pass as far as i can see.

If you want or feel ready for it, we can transfer this to the org. I already send you a transfer request, just in case :)
If we transfer it, we will need to change the gitmodules url.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "projects/svelte-add/adders/tailwindcss"]
 	path = projects/svelte-add/adders/tailwindcss
 	url = https://github.com/svelte-add/tailwindcss
+[submodule ".\\projects\\svelte-add\\adders\\github-pages"]
+	path = .\\projects\\svelte-add\\adders\\github-pages
+	url = https://github.com/manuel3108/svelte-add-github-pages.git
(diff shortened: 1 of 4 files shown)


babichjacob:
Yes, I'm going to leave this open for a little bit because it's clear `svelte-add` needs some internal infrastructure updates to support this.

I also tested out the adder and it worked well after I figured out I needed to configure the base path to be the same as the repository name. When I update this code, I'll include that as a question and include instructions that you have to make your app links like `"{dev ? '' : $paths.base}/about" for them to still work when deployed. There is a discussion / issue about how this is weird somewhere in the SvelteKit repository.

Will take care of these things later.

manuel3108:
Ok, thanks for testing that out. True, i forgot about the base-path thing. I only tested the svelte-kit skeleton template...

PR #249, opened 2023-01-05
Detect package manager based on lockfile

samrobbins85:
If you have an npm project but also have pnpm installed and use svelte add, you'll get a message to run `pnpm install`  to install new dependencies. This uses `detect-package-manager` to detect the project type based on the lockfile, so now it will tell the user to use `npm install`


babichjacob (approved):

manuel3108 (changes requested):
Hey @samrobbins85 

thanks for your PR!
Your changes work great, especially for non-interactive commands.

But I think that it currently does not work for interactive commands, that are executed by `npx svelte-add@latest`. Would you mind having a look at it?

I think we might need to add something like this, somewhere around here: 
https://github.com/svelte-add/svelte-add/blob/cc097553c91f6150973c842707387e7b5aa73ece/projects/svelte-add/index.js#L424

```js
defaultPackageManager = passedPackageManager ?? defaultPackageManager;
```

manuel3108:
Thanks for your contribution. I have added a few fixes that occurred during my testing. See #275

Issue #228, opened 2022-08-08
yields TODO: explain this error


GaryB432:
running `npm create @svelte-add/kit@latest app --with typescript+scss` yields `TODO: explain this error. debug info: output [ 'app2', 'typescript+scss' ] with undefined'] with undefined`

Windows 11

SMEETT:
Same problem here.

manuel3108:
For anyone in the future encountering this in the future, you are required to execute your command like this
```
npm create @svelte-add/kit@latest app -- --with typescript+scss
```
instead of this 
```
npm create @svelte-add/kit@latest app --with typescript+scss
```
Please not the two added dashes (--) in the first command, before passing any additional command line arguments.

Sadly this is a limitation of NPM, see here in example: [[Sending command line arguments to npm script](https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script)](https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script)

I'm going to create a PR for this in the near future.

GaryB432:
using 

```
npm create @svelte-add/kit@latest app -- --with typescript+css
```

does not give the error from the original issue.  However it creates both a `vite.config.js` and a `vite.cofig.ts` (both a javascript and typescript version)

manuel3108:
I have pushed a commit and released it with a more meaningful error message. 

Also, the other error you mention got fixed at the same time, see #234

PR #270, opened 2023-05-01
🐛 fix: update referenced packages (Update referenced package versions #261 create-svelte outdated #263)

### LINKED ISSUE #261: Update referenced package versions
In #259 we have seen that we are using quite a bunch of old versions for the packages that we reference, and that this can cause problems.

See here for the current versions:
https://github.com/svelte-add/svelte-add/blob/062154a5f6e723987a4d8b69b3894965b7869436/projects/svelte-add/package-versions.js#L2-L23

I have begun working out some new versions, but currently focused only on the ones with patch version changes. 
image
Everything with minor version changes and upwards should probably be analyzed and tested more deeply.

### LINKED ISSUE #263: create-svelte outdated
in https://github.com/svelte-add/svelte-add/blob/062154a5f6e723987a4d8b69b3894965b7869436/projects/create-kit/package.json#L27 you are still using v2, latest has been on v3 for a while now. 

To ensure users getting a current setup, please consider updating

manuel3108test:
resolves #261 
resolves #263


manuel3108:
I'm currently waiting to merge this, to see if we introduced any regressions with the last release. Just to be sure to be able to differentiate potential bugs from different features.

manuel3108:
I'm going to merge this either way now, as without this, other problems occur. See #228

PR #282, opened 2023-06-02
Create Pico CSS adder

### LINKED ISSUE #253: Add adder for Pico CSS.
Adding [Pico CSS](https://picocss.com/) to SvelteKit is relatively simple, but the ideal method is not obvious. It involves adding an id to app.html and setting a SASS variable:

> Pico depends on `main`, `footer`, `header` being direct children of `body` for certain layouts, but that's not possible with sveltekit if you are doing it right.
> 
> However, pico has a `$semantic-root-element` variable that you can set as this element: `<div style="display: contents">%sveltekit.body%</div>`. So far this has fixed that for me but I haven't tried everything.
> 
> I did it like this.
> 
> ```
> 
> <div id="pico-root" style="display: contents">%sveltekit.body%</div>
> ```
> 
> ```css-scss
> // variables.scss
> $semantic-root-element: "div#pico-root"
> 
> // OR whever you import pico
> $semantic-root-element: "div#pico-root"
> @import '@picocss/pico/scss/pico';
> ```
source: https://github.com/picocss/pico/issues/201#issuecomment-1398004719

I think app.html could be left intact if the proper selector is used. Maybe something like:

    $semantic-root-element: "body div:first-child"

~~Do +layout.svelte files add extra layers of  `<div>` containers that would cause this method to break? (Only a small part of Pico CSS depends on the `semantic-root-element`; some vertical spacing between container elements.)~~

**Edit:**  Inspecting https://learn.svelte.dev/tutorial/layouts seems to confirm layouts don't add additional layers of `<div>` containers. (Although more complex custom layouts with CSS grid, flex, etc may result in breaking Pico's semantic root.)

caio-bernardo:
Closes #253 

This changes are complementary to the [adder](https://github.com/caio-bernardo/picocss), which will be integrated on the svelte-add repo structure.


manuel3108:
Resolves #253 

Thanks for the PR!

Just tested it and it works flawlessly for Kit and Vite projects!

Sadly, I was not able to accept the transfer request, because I did not receive one. So I just took your repo and pushed it to: https://github.com/svelte-add/picocss. Hope that works out for you. 

The only thing is that it did not correctly detect the submodule from me. Here is what I did to resolve this:
* Undo your changes in the `.gitmodules` file
* Run `git submodule add https://github.com/svelte-add/picocss`

If you could do that here, that would be great.

Note: Don't care about the failing pipeline, I just saw that this is something I created yesterday.

caio-bernardo:
> Sadly, I was not able to accept the transfer request, because I did not receive one. So I just took your repo and pushed it to: https://github.com/svelte-add/picocss. Hope that works out for you.

Well, this is unexpected, maybe I did something wrong on the transfer request. Anyway, I'll make the changes right now.

manuel3108:
I found two small issues, but I was able to solve them by myself. 
* The git submodule was added to the wrong folder (or better: my instructions for adding it were wrong 👍 )
* There was a small typing typo here: https://github.com/svelte-add/svelte-add/pull/283/commits/3c8525a415a60d98ef2edee5ebcb8c2489c3fb17

Both things are resolved now, and the adder was published wit the latest update. Thanks for your work!

caio-bernardo:
I imagined there was something wrong with the git submodule, but the typo was a surprise. I'm glad I could contribute to this great project and hope to help in the future!

PR #290, opened 2023-06-24
🧹 chore: update package versions

manuel3108:


manuel3108:
Waiting for mdsvex support, see https://github.com/pngwn/MDsveX/pull/521

manuel3108:
Nahh, let's not wait. This issue will be automatically solved once mdsvex is ready, without an interaction necessary

PR #322, opened 2023-12-10
fix: overwrite caused while installing prettier-plugin-tailwindcss

### LINKED ISSUE #318: Svelte-add removes `prettier-plugin-svelte` and replaces it with `prettier-plugin-tailwindcss`
Hi.

First off, thanks for including `prettier-plugin-tailwindcss`. It was a chore doing it yourself.

That said, with the new release(s), using `svelte-add` to add tailwind, swaps plugins, and as a result, when I hit:

```
pnpm lint
```
I get the error
```bash
[error] Couldn't resolve parser "svelte".
 ELIFECYCLE  Command failed with exit code 1.
```

## Steps to reproduce:
1.
```
pnpm create svelte@latest test
```
2.
```
cd test && code .
```
Check your `.prettierrc` file. It should look like this
```
{
	"useTabs": true,
	"singleQuote": true,
	"trailingComma": "none",
	"printWidth": 100,
	"plugins": ["prettier-plugin-svelte"],
	"overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
}
```
4.
```
pnpx svelte-add@latest tailwindcss
```
Check your `.prettierrc` file again. It will look like this
```diff
{
	"useTabs": true,
	"singleQuote": true,
	"trailingComma": "none",
	"printWidth": 100,
-	"plugins": ["prettier-plugin-svelte"],
+	"plugins": ["prettier-plugin-tailwindcss"],
	"overrides": [
		{
			"files": "*.svelte",
			"options": {
				"parser": "svelte"
			}
		}
	]
}
```
Also, you can inspect your `package.json`. It effectively removes the `prettier-plugin-svelte`.

## Expectations
I want to have both plugins, so that in the end, it looks like this.
```diff
{
	"useTabs": true,
	"singleQuote": true,
	"trailingComma": "none",
	"printWidth": 100,
-	"plugins": ["prettier-plugin-svelte"],
-	"plugins": ["prettier-plugin-tailwindcss"],
+	"plugins": ["prettier-plugin-svelte", "prettier-plugin-tailwindcss"],
	"overrides": [
		{
			"files": "*.svelte",
			"options": {
				"parser": "svelte"
			}
		}
	]
}
```
---

I am more than happy to submit a P.R, if someone were so kind as to point me in the right direction. 🙂 

---
P.S. I am aware of this: https://github.com/svelte-add/svelte-add/issues/229, but having only one plugin breaks.
I think what should be inferred from Tailwind, https://github.com/tailwindlabs/prettier-plugin-tailwindcss#compatibility-with-other-prettier-plugins, is that they must be used at the same time, the only condition being that tailwind's plugin comes last.
If I'm missing something, or misunderstanding another, please let me know.

kakashy:
Please consider the changes made while attempting to install `prettier-plugin-tailwindcss`. I feel like all that's required is to have it as the tail of the plugins list in the `.prettierrc` file.
Should be able to fix #318 if successful


manuel3108:
Hey @kakashy 

thanks for your work. Could you also add a PR against the svelte-add/tailwind repository please?

Currently the only thing I am able to see is that you updated the submodule reference. But currently I cannot see the actual changes you have implemented.

Thanks for your help!

kakashy:
Hey @manuel3108 
Thanks for the reply. I'm new to sub-modules as well so sorry for any inconveniences caused. I'll look into how to do that.

manuel3108:
No problem. If there are any questions just hit me up

manuel3108:
Im on my phone so I would also need to Google it. But the easiest was for you would be to fork the repo you mentioned and just make the changes there. Then create a PR against that repo and leave everything else up to me.

I will migrate away from the submodules at some point, but it's going to be a while....

PR #326, opened 2023-12-16
fix: import `vitePreprocess` from `vite-plugin-svelte` in sveltekit

BlankParticle:
As of sveltekit v2 `vitePreprocess` is no longer exported from `@sveltejs/kit/vite`, So `vitePreprocess` always is imported from `@sveltejs/vite-plugin-svelte`.

This fixes the bug where `vitePreprocess` is imported 2 times


manuel3108:
@BlankParticle Thanks for your help. This has just been published!

PR #329, opened 2024-01-22
Added bun support

ThatXliner:


manuel3108:
Hey @ThatXliner
were you able to test this? Windows does not want to install WSL correctly.
From the other source code, it look like there might be other places that might need to be adjusted to make this work.

RubenPX:
@manuel3108 @ThatXliner 

 It's not required to use WSL anymore. [Bun is compatible with windows now](https://bun.sh/blog/bun-v1.1)

To install bun in windows, go to https://bun.sh ant it seays how to install it with powershell

manuel3108:
@RubenPX Thanks for your help in identifying this. 

`svelte-add` will be deprecated pretty soon, so it doesn't make sense to implement the changes here. I have already opened an issue in it's successor project `svelte-compose`

Electroid:
@manuel3108 Understand this package will soon be deprecated. It would be great to sneak #333 in before you do that, given that `svelte-add` is used in many places and will still be depended on for a while.

ThatXliner:
In that case, I will close my PR

PR #343, opened 2024-05-06
fix: install the correct `+layout.svelte` file for Tailwind adder

AdrianGonz97:
For the tailwind adder, `+layout.js` was being installed instead of the needed `+layout.svelte` file.

### DIFF
--- a/.changeset/cool-socks-punch.md
+++ b/.changeset/cool-socks-punch.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/tailwindcss": patch
+---
+
+fix: installs proper layout component file
--- a/.changeset/cool-socks-punch.md
+++ b/.changeset/cool-socks-punch.md
@@ -1,5 +1,0 @@
----
-"@svelte-add/tailwindcss": patch
----
-
-fix: installs proper layout component file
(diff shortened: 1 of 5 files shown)


manuel3108:
Thanks for your contribution! 

Could you enlighten me under which circumstances this was not working properly? Because I'm pretty sure this was working this weekend and the last pipeline runs also seem successful.

AdrianGonz97:
Sure, here's a base repro as a starting point (just a basic `create-svelte` template): https://github.com/AdrianGonz97/svelte-add-tailwind-layout-repro

2. `pnpm install`
3.  `pnpm dlx @svelte-add/tailwindcss@latest --typography false `

Notice how it will import the stylesheet in `src/routes/+layout.js` instead of a layout component file.

manuel3108:
Thanks for your input!

I'm honestly not sure what is the correct way to implement this. I tried searching for best practices inside the docs, but I was unable to do find anything. Based on that I have create an issue in the official `kit` repo in the hope of getting some good answers there: https://github.com/sveltejs/kit/issues/12202

manuel3108:
Based on the feedback we got, it looks like you made the right call!

I have adapted the other adders as well, and I'm going to merge once the tests pass!

PR #344, opened 2024-05-06
fix: fixed spacing for `package.json`

AdrianGonz97:
After execution, the `package.json` would be completely collapsed without spacing like so:

 ```json
{"name":"toolbar-test","version":"0.0.1","private":true,"scripts":{"dev":"vite dev","build":"vite build","preview":"vite preview","check":"svelte-kit sync && svelte-check --tsconfig ./tsconfig.json","check:watch":"svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch"},"devDependencies":{"@sveltejs/adapter-auto":"^3.0.0","@sveltejs/kit":"^2.0.0","@sveltejs/vite-plugin-svelte":"^3.0.0","autoprefixer":"^10.4.16","postcss":"^8.4.33","svelte":"^4.2.7","svelte-check":"^3.6.0","tailwindcss":"^3.4.1","tslib":"^2.4.1","typescript":"^5.0.0","vite":"^5.0.3"},"type":"module","dependencies":{"bits-ui":"^0.21.7","clsx":"^2.1.1","tailwind-merge":"^2.3.0","tailwind-variants":"^0.2.1"}}
```

This PR adds it back using tabs

### DIFF
--- a/.changeset/beige-avocados-kick.md
+++ b/.changeset/beige-avocados-kick.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/core": patch
+---
+
+fix: `package.json` spacing
--- a/packages/core/adder/execute.ts
+++ b/packages/core/adder/execute.ts
@@ -128,7 +128,7 @@ export async function installPackages(config: InlineAdderConfig<OptionDefinition
         }
     }
 
-    const packageText = await format(workspace, commonFilePaths.packageJsonFilePath, JSON.stringify(content));
+    const packageText = await format(workspace, commonFilePaths.packageJsonFilePath, JSON.stringify(content, null, "\t"));
     await writeFile(workspace, commonFilePaths.packageJsonFilePath, packageText);
 }
 


manuel3108:
This probably happened when creating a completely new project and applying an adder immediately, or was there another use case?

AdrianGonz97:
Not quite sure by what you mean by 'immediately'. Following the same steps from the [other PR](https://github.com/svelte-add/svelte-add/pull/343#issuecomment-2097435341) should yield the same results. 

For posterity, I'll add them here too:
2. `pnpm install`
3.  `pnpm dlx @svelte-add/tailwindcss@latest --typography false`
4. Select `pnpm` for the dependency install option

The `package.json` should be spaceless and jumbled up at this point.

manuel3108:
Thanks for the repro. This behavior did not happen to me previously, as I was mainly testing inside the monorepo, where a prettier configuration file was found, which caused the file to be formatted correctly. 

I have just created #345 with minor improvements over your implementation. Based on your implementation every indentation was converted to tabs, even if the user decided to use spaces inside his repository.

PR #352, opened 2024-05-11
fix: add slot element for layouts

AdrianGonz97:
Generated `+layout.svelte` files were missing their corresponding `slot` elements. Not quite sure how I missed this during the fix of #343 🤦‍♂️


manuel3108 (approved):

manuel3108:
Interesting, that the test's passed as well. Thanks for the fix!

PR #361, opened 2024-05-15
chore: Enable `strict` mode and adders install dependencies directly

### LINKED ISSUE #351: chore: Enable `strict` mode in `tsconfig.json`
### Topic

all packages

### Description

`strict` mode is useful for catching common mistakes with better guarantees of correctness. Enabling it also lights up the codebase like a christmas tree, so it'll take some finagling to fully implement it 😄

_No response_

### LINKED ISSUE #355: install dependencies directly, if lockfile present (without user input)
### Topic

core

### Description

As discussed in #307, if we detect that a lockfile is present (like `pnpm-lock.yaml` or similar), we can safely assume that's the package manager the user is using. Therefore we should skip asking the user which package manager he wants to use.

If we however cannot safely detect a lockfile, we should ask the user which package manager he would like to use. This is how it is implemented now. This will probably mainly be used after creating new projects.

Don't forget to think about monorepos and recursively checking the file path for a lockfile!

AdrianGonz97:
closes #351
closes #355

This PR enables TS's [strict mode](https://www.typescriptlang.org/tsconfig/#strict) and fixes several spots where the types were failing. I also took the liberty of fixing #355 as the logic had to be adjusted anyway to better satisfy the new type requirements.

Additional notes for this PR can be seen below 👇

### DIFF
--- a/packages/core/utils/prompts.ts
+++ b/packages/core/utils/prompts.ts
@@ -1,10 +1,17 @@
 import { cancel, intro, isCancel, outro, select, text, multiselect } from "@clack/prompts";
 
-export type PromptOption = {
-    label: string;
-    value: boolean | string;
-    hint?: string;
-};
+type Primitive = Readonly<string | boolean | number>;
+export type PromptOption<Value> = Value extends Primitive
+    ? {
+          value: Value;
+          label?: string;
+          hint?: string;
+      }
+    : {
+          value: Value;
+          label: string;
+          hint?: string;
+      };
 
 export function startPrompts(message: string) {
     intro(message);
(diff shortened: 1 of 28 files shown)


packages/core/utils/prompts.ts
```diff
@@ -1,10 +1,17 @@
...
+    ? {
+          value: Value;
+          label?: string;
+          hint?: string;
+      }
+    : {
+          value: Value;
+          label: string;
+          hint?: string;
+      };
```
manuel3108, line 14: Am I not seeing correctly, or is the only difference between the two types the optional label? Shouldn't the first type already include the second type?

Why are you declaring it explicitly?

manuel3108 (approved):
LGTM, there is just one minor thing where I'm not sure if I understand that correctly.

packages/core/utils/prompts.ts
AdrianGonz97: This type was sourced directly from `@clacks/prompts` as they don't expose it (for some reason 🤷): https://github.com/natemoo-re/clack/blob/45ee73bf33b25f9a8c7e1bb6117ccc165478bf4d/packages/prompts/src/index.ts#L208-L212

Without it, the types four our prompt wrappers aren't satisfied

PR #362, opened 2024-05-15
chore: slightly speed up tests by using `pnpm` instead of `npm`

AdrianGonz97:
This PR is a bit of an experiment to see if using `pnpm` is faster for tests than `npm`. It's noticeably faster on my machine (most of the time, using `npm` just hangs and it takes ages to complete), however I'm curious to see the effects of it in CI.


AdrianGonz97:
improvements in CI is negligible, if not worse. It's also causing the tests for bulma to fail - closing

manuel3108:
Yeah, I tried many thinks to make it faster as well, but I never managed to find something that works.

PR #365, opened 2024-05-16
chore: correctly type annotate `tailwind.config.ts`'s config when using typescript

### LINKED ISSUE #348: Correctly type annotate the config for the `tailwind.config.ts`
### Topic

tailwindcss

### Description

When using the tailwindcss adder while also having typescript installed, the adder will create a `tailwind.config.ts` file. This is normally fine, however the type annotation that's added for the config uses JSDoc, which won't correctly annotate the config object.

To reproduce:
1. `pnpm install`
1. `pnpm dlx @svelte-add/tailwindcss@latest --typography false`
1. Select `pnpm` for the dependency install option
1. Go to `tailwind.config.ts` and try to get autocompletions to trigger when editing the config object (it won't work)

There are a couple of ways to fix this.

Normal type annotation:
```ts
import type { Config } from "tailwindcss";

const config: Config = { ... }

export default config;
```

Using `as` or `satisfies`:
```ts
import type { Config } from "tailwindcss";

export default { ... } as Config;

// or...
export default { ... } satisfies Config;
```

Or we can just keep it as a `.js` file. However, type-saftey won't be enforced unless the file includes a `// @ts-check` at the top, or if the file is included in the `includes` field of the `tsconfig.json`.

AdrianGonz97:
closes #348 

Went with the `as Config` route as it was the easiest to implement with relatively low risk of causing side-effects down the line.

### DIFF
--- a/packages/ast-manipulation/js/imports.ts
+++ b/packages/ast-manipulation/js/imports.ts
@@ -35,7 +35,7 @@ export function addDefault(ast: AstTypes.Program, importFrom: string, importAs: 
     addImportIfNecessary(ast, expectedImportDeclaration);
 }
 
-export function addNamed(ast: AstTypes.Program, importFrom: string, exportedAsImportAs: Record<string, string>) {
+export function addNamed(ast: AstTypes.Program, importFrom: string, exportedAsImportAs: Record<string, string>, isType = false) {
     const specifiers = Object.entries(exportedAsImportAs).map(([key, value]) => {
         const specifier: AstTypes.ImportSpecifier = {
             type: "ImportSpecifier",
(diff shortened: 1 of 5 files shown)


packages/ast-manipulation/js/imports.ts
```diff
@@ -35,7 +35,7 @@ export function addDefault(ast: AstTypes.Program, importFrom: string, importAs:
     addImportIfNecessary(ast, expectedImportDeclaration);
 }
 
-export function addNamed(ast: AstTypes.Program, importFrom: string, exportedAsImportAs: Record<string, string>) {
+export function addNamed(ast: AstTypes.Program, importFrom: string, exportedAsImportAs: Record<string, string>, isType = false) {
```
AdrianGonz97, line 38: I went with adding a new optional parameter to this existing function as the implementation of adding a type import vs a regular import were nearly identical.

If it's clearer to make a new function instead (e.g. `addNamedType`), I'm open to make that change as well

manuel3108: I think that's a good decision

manuel3108 (approved):

PR #376, opened 2024-05-26
feat: add `--default` flag

### LINKED ISSUE #350: Adjust the opt-out behavior of CLI flags
### Topic

applies to all adders

### Description

It would be preferable if the CLI flags were designed to be **opt-in** rather than **opt-out**.

One can imagine that when a couple of more plugins are added, it would quickly become cumbersome to add all of these opt-out flags just to avoid the CLI wizard:

```
npx @svelte-add/tailwindcss@latest --typography false --forms false --scrollbars false --other-plugin false --one-last-one false
```

Instead, it should be the inverse, where adding the flag includes the addon/plugin:
```
npx @svelte-add/tailwindcss@latest --typography --forms
```

### Alternative

An alternative could be to include a `--none` flag to preserve the default state of running through the CLI wizard, but also not having to add all of the flags to avoid it either:
```
npx @svelte-add/tailwindcss@latest --none
```

AdrianGonz97:
Closes #350

This PR adds the `--default` flag as a shorthand to install the default adder options for **missing options**.

For example, say that we have 2 tailwind options `typography` and `forms` (assume both have their defaults set to `false`). If we were to run the command like so:
```bash
@svelte-add/tailwindcss --default --typography
```
the `typography` plugin will still be added as it was manually specified, but the `forms` plugin will not.

### DIFF
--- a/packages/core/adder/execute.ts
+++ b/packages/core/adder/execute.ts
@@ -74,7 +74,7 @@ export async function executeAdders<Args extends OptionDefinition>(
 
     let workingDirectory: string | null;
     if (isTesting) workingDirectory = remoteControlOptions.workingDirectory;
-    else workingDirectory = await determineWorkingDirectory(commonCliOptions.path);
+    else workingDirectory = determineWorkingDirectory(commonCliOptions.path);
     workingDirectory = await detectSvelteDirectory(workingDirectory);
     const createProject = workingDirectory == null;
     if (!workingDirectory) workingDirectory = process.cwd();
@@ -130,6 +130,16 @@ async function executePlan<Args extends OptionDefinition>(
     // preconditions
     if (!executionPlan.commonCliOptions.skipPreconditions)
         await validatePreconditions(adderDetails, executingAdder.name, executionPlan.workingDirectory, isTesting);
+
+    // applies the default option value to adder's cli options
+    if (executionPlan.commonCliOptions.default) {
+        for (const adder of adderDetails) {
+            const adderId = adder.config.metadata.id;
+            for (const [option, value] of Object.entries(adder.config.options)) {
+                executionPlan.cliOptionsByAdderId[adderId][option] = value.default;
+            }
+        }
+    }
 
     // ask the user questions about unselected options
     await requestMissingOptionsFromUser(adderDetails, executionPlan);
--- a/packages/core/adder/execute.ts
+++ b/packages/core/adder/execute.ts
@@ -131,12 +131,12 @@ async function executePlan<Args extends OptionDefinition>(
     if (!executionPlan.commonCliOptions.skipPreconditions)
         await validatePreconditions(adderDetails, executingAdder.name, executionPlan.workingDirectory, isTesting);
 
-    // applies the default option value to adder's cli options
+    // applies the default option value to missing adder's cli options
     if (executionPlan.commonCliOptions.default) {
         for (const adder of adderDetails) {
             const adderId = adder.config.metadata.id;
             for (const [option, value] of Object.entries(adder.config.options)) {
-                executionPlan.cliOptionsByAdderId[adderId][option] = value.default;
+                executionPlan.cliOptionsByAdderId[adderId][option] ??= value.default;
             }
         }
     }
--- a/packages/core/adder/options.ts
+++ b/packages/core/adder/options.ts
@@ -63,7 +63,7 @@ export const availableCliOptions: AvailableCliOptions = {
         processedCliArg: "default",
         type: "boolean",
         default: false,
-        description: "Installs default adder options",
+        description: "Installs default adder options for unspecified options",
         allowShorthand: true,
     },
     path: {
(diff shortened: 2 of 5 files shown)


packages/core/adder/options.ts
```diff
@@ -37,8 +37,9 @@ export type OptionValues<Args extends OptionDefinition> = {
             : never;
 };
 
-export type AvailableCliOptionKeys = "path" | "skipPreconditions" | "skipInstall";
+export type AvailableCliOptionKeys = keyof AvailableCliOptionKeyTypes;
```
AdrianGonz97, line 40: Simplified this so we don't have to declare the names twice

packages/core/utils/dependencies.ts
```diff
@@ -30,7 +30,6 @@ export async function suggestInstallingDependencies(workingDirectory: string) {
     }
 
     if (!selectedPm || !packageManagers[selectedPm]) {
-        console.log("Skipped installing dependencies");
```
AdrianGonz97, line 33: Removed this log for two reasons: 
- it wasn't using clack's logger so the alignment was off
- it's also redundant

The latter reason is really the core reason to remove it
img
the prompt with the response already conveys that we don't want to install deps, so there's no need to say that we're _not_ doing something

packages/core/adder/options.ts
manuel3108: I knew I was missing something while creating those types. Thanks for catching that.

packages/core/utils/dependencies.ts
manuel3108: Great catch!

packages/core/adder/execute.ts
```diff
@@ -74,7 +74,7 @@ export async function executeAdders<Args extends OptionDefinition>(
 
     let workingDirectory: string | null;
     if (isTesting) workingDirectory = remoteControlOptions.workingDirectory;
-    else workingDirectory = await determineWorkingDirectory(commonCliOptions.path);
+    else workingDirectory = determineWorkingDirectory(commonCliOptions.path);
```
manuel3108, line 77: Just out of interest: Is this something a tool highlighted to you or something you found out by accident?

Coming mainly from C#, my IDE and even the compiler would have yelled at me for this.

packages/core/adder/options.ts
```diff
@@ -57,6 +58,14 @@ export type AvailableCliOption = {
 export type AvailableCliOptions = Record<AvailableCliOptionKeys, AvailableCliOption>;
 
 export const availableCliOptions: AvailableCliOptions = {
+    default: {
+        cliArg: "default",
+        processedCliArg: "default",
+        type: "boolean",
+        default: false,
+        description: "Installs default adder options",
```
manuel3108, line 66: Should we add something to the description, indicating that `--default` can be used in combination with other adder options, like you did in the PR description?

manuel3108 (approved):
Looks good and works as expected.

Although where I'm not 100% sure about, so let me just think out loud.
The goal of #350 is to minimize the command length and the required selections within the cli. 

Consider this command:
```sh
npx @svelte-add/tailwindcss --default
```

As of now, this interaction might create two required interactions for the user
* At least one precondition fails
* We could not detect a package manager

So the question arises if `--default` should implicitly add `--skip-preconditions` and `--skip-install`?

On the one side we could argue that with that change we could guarantee the user a zero interaction setup, but he would be missing a lot of the benefits. On the other side, we already only displays these prompts if really required (> 1 failed precondition, no PM detected) so a lot of these cli commands should already be interaction free.

Based on that I would think it's the best to leave the implementation as-is. What do you think?

packages/core/adder/execute.ts
AdrianGonz97: Typescript's LSP should provide a warning about it. Here's what it looks like in vscode, although it's quite subtle:
img

---

We could add the [`@typescript-eslint/eslint-plugin`](https://typescript-eslint.io/) plugin with it's recommended typed defaults (which are pretty solid, IMHO) which has a rule to catch these particular cases:
img

packages/core/adder/options.ts
AdrianGonz97: We absolutely could! Would `Installs default adder options for unspecified options` suffice, or would that be _too_ vague?

AdrianGonz97:
> So the question arises if `--default` should implicitly add `--skip-preconditions` and `--skip-install`?
> 
> On the one side we could argue that with that change we could guarantee the user a zero interaction setup, but he would be missing a lot of the benefits. On the other side, we already only displays these prompts if really required (> 1 failed precondition, no PM detected) so a lot of these cli commands should already be interaction free.
> 
> Based on that I would think it's the best to leave the implementation as-is. What do you think?

I think it should be left as-is. `--default` should really only affect the adder's specific options rather than the whole experience.

packages/core/adder/execute.ts
manuel3108: That's interesting. I have never seen something like this in vscode. Good catch, and good suggestion. I will add that eslint plugin later.

packages/core/adder/options.ts
manuel3108: I think that's better than what we have right now and I don't have any better wording to suggest. I have just update the text.

manuel3108:
Cool, then we agree!

PR #378, opened 2024-05-28
chore: enforce stricter eslint rules

manuel3108:
In #376 we detected that our current setup was missing a lot of relevant use-cases, such as async await detection.

I have configured the project according to these docs: https://typescript-eslint.io/getting-started/ and already fixed about 100 warnings or so. But those were the easiest ones. I'm out of time to continue working on this now.

- [x] fix all those warnings
- [x] consider if we are able to solve all those warnings, especielly the `any` related warnings, or if we need to disable this rule globally as a last resort.
- [ ] ~~re-enable checking of the website project (currently caused problems because of the appropriate ´tsconfig.json` / project currently ignored by rules)~~ (see comment below)
- [x] consider using `strictTypeChecked` instead of `recommendedTypeChecked`


AdrianGonz97:
> consider using `strictTypeChecked `instead of `recommendedTypeChecked`

tbh, I think `recommendedTypeChecked` is good enough for our use-case. I'm not that big of a fan of eslint rules that hold strong opinions over _the logic_ of the code, so `strictTypeChecked` is a bit too opinionated for my taste. For instance, `@typescript-eslint/no-dynamic-delete` feels a bit silly

packages/core/adder/categories.ts
```diff
@@ -3,9 +3,10 @@ export type CategoryInfo = {
     name: string;
     description: string;
 };
-
+export type CategoryKeys = "styling" | "tools" | "asd";
```
AdrianGonz97, line 6: this `"asd"` literal seems to have been accidentally left in :p

packages/testing-library/utils/test-cases.ts
```diff
@@ -160,8 +161,12 @@ export async function runTestCases(testCases: Map<string, TestCase[]>, testOptio
...
     });
 
-    const rejectedAsyncPromisesResult = allAsyncResults.rejectedIndexes.map<AdderError>((x) => allAsyncResults.taskResults[x]!);
-    const rejectedSyncPromisesResult = allSyncResults.rejectedIndexes.map<AdderError>((x) => allSyncResults.taskResults[x]!);
+    const rejectedAsyncPromisesResult = allAsyncResults.rejectedIndexes.map<AdderError>(
+        (x) => allAsyncResults.taskResults[x] as unknown as AdderError,
+    );
+    const rejectedSyncPromisesResult = allSyncResults.rejectedIndexes.map<AdderError>(
+        (x) => allSyncResults.taskResults[x] as unknown as AdderError,
+    );
```
AdrianGonz97, line 169: i think this is another scenario where `strictTypeChecked` is being too restrictive again - preventing the type assertion with the `!` and forcing us to use the `as unknown as Type` hack

packages/core/adder/categories.ts
manuel3108: Thanks! I have also properly fixed the error that I had while accidentally adding that!

packages/testing-library/utils/test-cases.ts
manuel3108: There is an argument to be made about code clarity vs. readability, as you said. In this case I totally agree with the strict rule, if we do things that make typescript unhappy, we should at least be able to identify them quickly. 

Even when I got this error and the error message stated something about an exclamation mark, i read past it because it's so small. Let's keep it this way.

manuel3108:
I was under the same impression you mentioned here before trying it out. After seeing the errors that came up, I decided to commit on `strict` mode as many of them were useful. 

As you said, some of those rules are questionable, but overall I think enabling `strict` mode is an improvement. And if we don't agree with some rules, we always have the possibility to disable them. In fact, that's what I have done with two rules already.

```
"@typescript-eslint/no-unnecessary-condition": "off",
"@typescript-eslint/no-dynamic-delete": "off",
```

manuel3108:
This PR is ready, but I'll wait for #379 to avoid merge conflicts there.

manuel3108:
Blocked by #382

packages/core/utils/common.ts
```diff
@@ -62,6 +67,7 @@ export async function executeCli(
             if (code == 0) {
                 resolve(undefined);
             } else {
+                // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors
                 reject(errorText);
```
benmccann, line 71: I think all you need to do to fix this is:
```suggestion
                reject(new Error(errorText));
```

packages/testing-library/utils/test.ts
```diff
@@ -13,8 +13,9 @@ export async function runTests(page: Page, adder: AdderWithoutExplicitArgs, opti
         click: async (selector, waitForNavigation) => {
             await click(page, selector, waitForNavigation);
         },
+        // eslint-disable-next-line @typescript-eslint/require-await
```
benmccann, line 16: why remove the `await` if it causes you to add this?

packages/website/src/lib/adder.js
```diff
@@ -9,7 +9,7 @@ import { groupBy } from "@svelte-add/core/internal";
 
 /**
  *
```
benmccann, line 11:
```suggestion
```

packages/website/src/lib/adder.js
```diff
@@ -9,7 +9,7 @@ import { groupBy } from "@svelte-add/core/internal";
...
- * @param {string | undefined} [category]
+ * @param {string} [category]
  * @returns
```
benmccann, line 13:
```suggestion
```

package.json
```diff
@@ -45,6 +45,7 @@
         "rollup": "^4.18.0",
         "rollup-plugin-add-shebang": "^0.3.1",
         "rollup-plugin-preserve-shebangs": "^0.2.0",
-        "typescript": "^5.4.5"
+        "typescript": "^5.4.5",
+        "typescript-eslint": "^7.11.0"
```
benmccann, line 49: you might want to use the alpha version of v8. version 7 isn't officially supported with eslint 9. and v8 has a different set of recommended rules, so if we're starting fresh it might be better to target those. you try could removing the ignored rules here and see if they're already removed from the recommended set in v8

packages/testing-library/utils/test.ts
manuel3108: Because `expectUrlPath` is non async function, and the await was misplaced previously. 

If I also remove the `async` keyword from this `expectUrlPath: async (path) => {` the types are also not happy because the tests should be an async function. Is there any other way to solve this?

package.json
manuel3108: Good hint. Also tested the new rules, but the two rules bothering us are still present in the alpha.

manuel3108:
Ignoring the refactoring of the website types for now, as we don't exactly know if this site will still be required after rewrite of the official svelte site.

Issue #338, opened 2024-05-05
Tauri


manuel3108:
### Website

a

### Documentation

a

### Integration summary

1. install package `tool`
2. modify `svelte-config.js` and to add tool as preprocessor
3. import tool inside `+layout.svelte`
4. and so on...

PR #379, opened 2024-05-29
feat: Drizzle adder

### LINKED ISSUE #298: Add Drizzle
[drizzle-orm](https://github.com/drizzle-team/drizzle-orm)

AdrianGonz97:
Closes #298

Here are the implemented options for each DB provider:
- SQLite
	- `better-sqlite3` (default)
	- `@libsql/client` general purpose client
	- `turso` is the `@libsql/client` but configured for Turso
- MySQL
	- `mysql2` (default)
	- `@planetscale/database` for Planetscale
- PostgreSQL
	- ~~`pg` (default)~~ removed as it uses CJS
	- `postgres` (default)
	- `@neondatabase/serverless` for Neon

The defaults are good general purpose clients while the rest are a bit more vendor specific (with the exception of `@libsql/client` which can also be used as a general purpose client too).

---

## Todo:
- [x] implement the premade `docker-compose.yml` as an additional option for `postgres` and `mysql`
- [x] add tests
	- [x] sqlite
	- [x] mysql
	- [x] postgres
- [x] fix website to consider the new Select question option


.github/workflows/changesets.yml
```diff
@@ -21,19 +20,31 @@ jobs:
               with:
                   fetch-depth: 0
 
+            - uses: pnpm/action-setup@v4
+              with:
+                  version: 9
```
benmccann, line 25: If you just add `"packageManager": "pnpm@9.3.0",` in the root `package.json` then you don't have to add `with: version: 9` so many places

AdrianGonz97: oh, neat! didn't know that was a feature of the action. thanks for the suggestion

manuel3108:
I'm not familiar with `drizzle` at all. Looking at some of the configs that were generated it looks like I played with it a few months ago, but that's all I can remember. This comment will only cover this PR from a user perspective. I will add a review later on.

---

Overall the setup seems to work as expected, as far as I can tell. I will add a few notes / potential issues below that can be discussed.

<details><summary>Cli output</summary>
<p>

┌  Welcome to @svelte-add/drizzle@1.0.0
│
◇  Create new Project?
│  Yes
│
◇  Where should we create your project?
│    (hit Enter to use 'temp')
│
◇  Which project type do you want to create?
│  SvelteKit
│
◇  Downloading initializer cli...
Need to install the following packages:
  create-svelte@6.3.0
Ok to proceed? (y)

create-svelte version 6.3.0

┌  Welcome to SvelteKit!
│
◇  Which Svelte app template?
│  Skeleton project
│
◇  Add type checking with TypeScript?
│  Yes, using JavaScript with JSDoc comments
│
◇  Select additional options (use arrow keys/space bar)
│  none
│
└  Your project is ready!

✔ Type-checked JavaScript
◇  Template initialized
│
◇  Preconditions: ───────────────────────────────────╮
│                                                    │
│  - clean working directory (Found modified files)  │
│                                                    │
├────────────────────────────────────────────────────╯
│
◇  Preconditions failed. Do you wish to continue?
│  Yes
│
◇  Which database would you like to use?
│  SQLite
│
◇  Which SQLite client would you like to use?
│  better-sqlite3
│
◇  Successfully installed dependencies
│
└  You're all set!

</p>
</details> 


1. Should we add a note with usage instructions, similar to the square with the preconditions? It could indicate the user that we added new scripts to `package.json` and that the user needs to run at least `db:push` to setup the database? (of course, after modifying the connection string, if required)
2. Should we add the possibility to add a small usage example? I would think of something like reading all users of the user table and displaying it in `index.svelte`? Because right now I would need to consider the drizzle docs to see on how to get started. Alternatively, we could point the user to the correct docs in the usage notes explained in 1.
3. running `db:migrate` leaves me with this error message (sqllite, better-sqllite): Error: Can't find meta/_journal.json file (see [here](https://www.answeroverflow.com/m/1093225882879017080))
4. #387 
5. MySQL: Local setup with docker is proposed and added to the `package.json` even if docker is not installed (I reinstalled my windows recently). Running `db:start` fails with a reasonabl command `'docker' is not recognized as an internal or external command, operable program or batch file.`. In my opinion this is totally fine and works as expected. Nevertheless I wanted to mention it here. 

I think that's everything I found when applying this adder as a user.

adders/drizzle/config/tests.js
```diff
@@ -0,0 +1,100 @@
...
+    docker: options.docker.default,
+};
+
+export const tests = defineAdderTests({
+    options,
+    optionValues: [
+        { ...defaultOptionValues, database: "sqlite", sqlite: "better-sqlite3" },
+        { ...defaultOptionValues, database: "sqlite", sqlite: "libsql" },
+        { ...defaultOptionValues, database: "mysql", mysql: "mysql2", docker: true },
+        { ...defaultOptionValues, database: "postgresql", postgresql: "postgres.js", docker: true },
```
manuel3108, line 17: Why do we have `defaultOptionValues`? Looks like we are re-defining all those options nevertheless. Or is it because we need to specify a valid option for `mysql` even if we are using `sqlite`?

adders/drizzle/config/tests.js
```diff
@@ -0,0 +1,100 @@
...
+    ],
+    runSynchronously: true,
```
manuel3108, line 19: Is there a reason for executing them synchronously? If yes, could you please add a comment indicating why this is the case similar to `storybook`?

adders/drizzle/config/tests.js
```diff
@@ -0,0 +1,100 @@
...
+
+                    function insertUser(${typescript.installed ? "value: typeof user.$inferInsert" : "value"}) {
+                        return db.insert(user).values(value);
+                    }
+                    `,
+                );
+            },
+        },
+        {
+            // override the config so we can remove strict mode
```
manuel3108, line 64: Why would this be required? Wouldn't it be simpler to just search and replace that part, instead of redefining the whole config?

adders/drizzle/config/tests.js
```diff
@@ -0,0 +1,100 @@
...
+                `;
+            },
+        },
+        {
+            name: () => "package.json",
+            contentType: "json",
+            content: ({ data }) => {
+                // executes after pnpm install
+                data.scripts["postinstall"] ??= "pnpm run db:push";
+            },
```
manuel3108, line 89: I would consider this more of a hack than a solution. What do you think?

Wouldn't it be better if we introduce pre and post test hooks?

packages/ast-manipulation/js/common.ts
```diff
@@ -57,3 +57,41 @@ export function addFromString(ast: AstTypes.BlockStatement | AstTypes.Program, v
...
+        throw new Error("value passed was not an expression");
+    }
+
+    return statement.expression;
+}
+
+export function statementFromString(value: string): AstKinds.StatementKind {
+    const program = parseScript(value);
+
+    return program.body[0];
```
manuel3108, line 74: In order to provide consistency with the method above, shouldn't we also check the generated statement type or throw an exception if anything other was found?

I think something like this should work:
```js
const program = parseScript(value);
const statement = program.body[0];
if (statement.type && statement.type.includes("statement")) {
    throw new Error("value passed was not an statement");
}

return statement.expression;
```

packages/ast-tooling/index.ts
```diff
@@ -44,7 +48,7 @@ export function parseScript(content: string): AstTypes.Program {
         },
     }).program;
 
-    return jsAst;
+    return stripAst(jsAst, "loc");
```
manuel3108, line 51: Can you explain why this was necessary? Is this just for better debug printing?

I was under the impression that `loc` should help the serialize to keep the code as close as possible to the original implementation that was parsed.

packages/core/adder/options.ts
```diff
@@ -1,40 +1,47 @@
...
 
+export type SelectQuestion<Value = any> = {
+    type: "select";
+    default: Value;
+    options: PromptOption<Value>[];
+};
+
 export type BaseQuestion = {
     question: string;
+    // TODO: we want this to be akin to OptionValues<Args> so that the options can be inferred
```
manuel3108, line 28: I assume this was overlooked. Or is it something you didn't get working?

packages/core/adder/preconditions.ts
```diff
@@ -94,6 +94,8 @@ export async function validatePreconditions<Args extends OptionDefinition>(
         }
     }
     if (preconditionLog.length > 0) {
+        if (isTesting) return;
```
manuel3108, line 97: Does not match with the changeset: `fix: preconditions are now ignored during tests`.
This does only skip the logging of the preconditions as far as i can tell.

packages/tests/index.js
```diff
@@ -57,3 +64,21 @@ async function getAdder(adderName) {
 
     return adder;
 }
+
+const cwd = path.resolve(fileURLToPath(import.meta.url), "..");
+
+// We're using `execSync` instead of our `executeCli` because we need the cleanup to be synchronous
+function startDocker() {
+    console.log("Starting docker containers");
+    execSync("docker compose up --detach", { cwd, stdio: "inherit" });
```
manuel3108, line 73: Anything that would prevent us from using `stdio: "pipe"`.

As of now the test outputs, especially in the GitHub pipeline, are completely cluttered with layer download information.

packages/tests/index.js
```diff
@@ -57,3 +64,21 @@ async function getAdder(adderName) {
...
+}
+
+function stopDocker() {
+    if (usingDocker) return;
+    console.log("Stopping docker containers");
+    execSync("docker compose down --volumes", { cwd, stdio: "inherit" });
```
manuel3108, line 79: see above

manuel3108 (changes requested):
Overall, everything looks really good, thanks for the fantastic work. Just some minor things here and there. 

Keep in mind I have not tested every possible combination at this point, but I'm pretty sure you have done so.

adders/drizzle/config/adder.js
```diff
@@ -0,0 +1,355 @@
+import { categories, defineAdderConfig, generateAdderInfo } from "@svelte-add/core";
+import pkg from "../package.json";
+import { options } from "./options";
+
+export const adder = defineAdderConfig({
+    metadata: {
+        ...generateAdderInfo(pkg),
+        name: "Drizzle",
+        description: "Headless TypeScript ORM with a head.",
```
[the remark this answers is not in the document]
AdrianGonz97: I nabbed this from their [gh repo's description](https://github.com/drizzle-team/drizzle-orm):
img

although in hindsight, perhaps `Headless ORM for NodeJS, TypeScript and JavaScript` would be a better description. this can be found right below the logo in the README

adders/drizzle/config/tests.js
AdrianGonz97: Indeed, it is a hack 😅 

Pre and post test hooks would be great in these cases

adders/drizzle/config/adder.js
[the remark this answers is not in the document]
manuel3108: Yeah, that sounds way better!

packages/ast-tooling/index.ts
AdrianGonz97: The `loc` prop was needed to be stripped as it was causing false negatives during node comparisons, which often resulted in broken code when an adder is ran twice on a project (i.e. it would re-add a line that already existed). 

Consider the following case:
```js
const foo = {
	bar: 'baz'
};
```

```js
const foo = {
	bar: "baz"
};
```

these two cases are functionally equivalent, but to the node comparator, they are completely different as `loc` keeps track of which quote variation it uses.

adders/drizzle/config/tests.js
manuel3108: Let's leave it like this for now, I have opened #389 for tracking

packages/ast-tooling/index.ts
manuel3108: Sounds logical, thanks for the explanation!

packages/core/adder/preconditions.ts
AdrianGonz97: thanks! moved it to the top of the function definition, which should prevent them from running.

packages/ast-manipulation/js/common.ts
AdrianGonz97: Trying it out, I don't think validating it will work. For instance, a variable declaration is a statement, but its `type` is `VariableDeclaration`, which doesn't contain the word `Statement`. Perhaps just checking if the value `program.body[0]` exists is good enough for this

adders/drizzle/config/tests.js
AdrianGonz97: Default values are needed to satisfy the type of `optionValues`. Since the type for `options` doesn't use a discriminated union, nor are they considered 'optional', each db type must be explicitly defined/initialized. In this case, we're just defining them with their default state, which is an explicit `undefined`.

adders/drizzle/config/adder.js
```diff
@@ -0,0 +1,355 @@
...
+            name: () => `docker-compose.yml`,
+            contentType: "text",
+            condition: ({ options }) => options.docker && (options.mysql === "mysql2" || options.postgresql === "postgres.js"),
+            content: ({ content, options }) => {
+                // if the file already exists, don't modify it
+                // (in the future, we could add some tooling for modifying yaml)
+                if (content.length > 0) return content;
+
+                const db = options.database === "mysql" ? "mysql" : "postgres";
+                const port = options.database === "mysql" ? "3306" : "5432";
```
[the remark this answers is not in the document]
AdrianGonz97: not sure i follow when you say: "extract them and make them file scoped" 😅

packages/core/adder/preconditions.ts
manuel3108: After further consideration: is this really what we want?

Let's assume #338 would be implemented already. Tauri requires rust and cargo to be installed on the dev's machine. Not running the preconditions will not make the dev aware that this is required.

But I can guess where you are coming from. Probably the `clean working directory (Found modified files)` is bothering you. Wouldn't it be enough if we disable that specific precondition during testing?

packages/core/adder/options.ts
AdrianGonz97: heh, yea I wasn't able to figure it out that day and never got around to fixing it. i'll take another crack at it in a few

packages/ast-manipulation/js/common.ts
manuel3108: Sound good!

packages/core/adder/preconditions.ts
AdrianGonz97: The problem isn't so much the logging, that's fine, it's just that they're causing the tests to fail when testing individual adders with `pnpm test drizzle`

AdrianGonz97:
> 1. Should we add a note with usage instructions, similar to the square with the preconditions? It could indicate the user that we added new scripts to `package.json` and that the user needs to run at least `db:push` to setup the database? (of course, after modifying the connection string, if required)

heh, you were just reading my mind! I think we should probably add some sort of a `nextSteps` hook that's executed after the adder is installed that provides the logger so that the adder can print follow-up instructions

> 2. Should we add the possibility to add a small usage example? I would think of something like reading all users of the user table and displaying it in `index.svelte`? Because right now I would need to consider the drizzle docs to see on how to get started. Alternatively, we could point the user to the correct docs in the usage notes explained in 1.

I considered adding one, but I wasn't sure how far we should really go in terms of having to maintain the example. For other adders, such as tailwind, we don't really provide a usage example for them AFAIK. I think pointing them to the docs via the `nextSteps` hook would be the best solution

> 3. running `db:migrate` leaves me with this error message (sqllite, better-sqllite): Error: Can't find meta/_journal.json file (see [here](https://www.answeroverflow.com/m/1093225882879017080))

i'll look into this, thanks!

> 5. MySQL: Local setup with docker is proposed and added to the `package.json` even if docker is not installed (I reinstalled my windows recently). Running `db:start` fails with a reasonabl command `'docker' is not recognized as an internal or external command, operable program or batch file.`. In my opinion this is totally fine and works as expected. Nevertheless I wanted to mention it here.

I think this is a fine tradeoff given that we don't have a "pre-question condition hook" for cases like these. If they're selecting `yes` for docker, I think we can (hopefully) assume they know what it is

manuel3108:
> heh, you were just reading my mind! I think we should probably add some sort of a `nextSteps` hook that's executed after the adder is installed that provides the logger so that the adder can print follow-up instructions

see #390 

> I considered adding one, but I wasn't sure how far we should really go in terms of having to maintain the example. For other adders, such as tailwind, we don't really provide a usage example for them AFAIK. I think pointing them to the docs via the `nextSteps` hook would be the best solution

True, completely forgot about that. Let's leave it as is for now.

adders/drizzle/config/adder.js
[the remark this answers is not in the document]
manuel3108: I was thinking of having global variables with the ports specified.

packages/core/adder/options.ts
AdrianGonz97: i changed it _a tiny bit_ so that it isn't just a flat out `any`, but I still couldn't figure this one out. i'm not sure if there's a way to derive the prop types of an object that the condition is a child of.

benmccann:
This is sick!!!

We probably don't need to print the preconditions if they passed

I'm a bit interested in integrating svelte-add into create-svelte. In that case, we'd need a way to skip the precondition of having a git repo

Should we put postgres before mysql? I feel like it's a little more popular with devs these days

Choosing the client felt a bit overwhelming to me and I didn't know how to choose. Perhaps we could give some more guidance.

For sqlite, maybe we can recommend one? Do we need libsql and better-sqlite3 or when would you choose one vs the other? Maybe we can drop ont?
- better-sqlite3 (recommended for local development)
- Turso (popular hosted platform)

For postgres:
- Postgres.JS (for traditional Node environments)
- @neondatabase/serverless (for serverless environments)

For mysql, do we even need multiple options?

AdrianGonz97:
> Should we put postgres before mysql? I feel like it's a little more popular with devs these days

sure! the current order was just the order i implemented them in

how does this order sound?
- PostgreSQL
- MySQL
- SQLite

> For sqlite, maybe we can recommend one? Do we need libsql and better-sqlite3 or when would you choose one vs the other? Maybe we can drop one?
> 
> * better-sqlite3 (recommended for local development)
> * Turso (popular hosted platform)

If we had to just choose one, then I would probably just keep `libsql` since it works as both, a general purpose client and a client for Turso, making the implementation of the adder a bit simpler.

> Choosing the client felt a bit overwhelming to me and I didn't know how to choose. Perhaps we could give some more guidance.

I can definitely see that and the thought did occur to me to provide better hints as guidance, though I wasn't sure of the most concise way to put it.

> For postgres:
> 
> * Postgres.JS (for traditional Node environments)
> * @neondatabase/serverless (for serverless environments)

I believe both clients are good for serverless environments, so I wouldn't want to distinguish them by that

maybe instead:
* Postgres.JS (for general purpose applications)
* @neondatabase/serverless (serverless client for Neon)

> For mysql, do we even need multiple options?

heh, I didn't want `mysql2` to be all alone, so I added planetscale. although, I don't know how popular they are nowadays

benmccann:
> how does this order sound?

Sounds good to me. I just checked state of db and that seems to match their overall popularity 

> If we had to just choose one, then I would probably just keep libsql since it works as both,

Sounds fine to me. I don't have an opinion on whether we limit to one or add more as long as there's some guidance 

> I believe both clients are good for serverless environments, so I wouldn't want to distinguish them by that

Yeah, that's probably right

Maybe something like:
- Postgres.JS (recommended for most users)
- Neon (popular hosted platform)

> I didn't want mysql2 to be all alone, so I added planetscale. although, I don't know how popular they are nowadays

Fair enough 😆 I think it's fine to have multiple as long as there's a clear default that people don't need to think about

manuel3108:
> I'm a bit interested in integrating svelte-add into create-svelte. In that case, we'd need a way to skip the precondition of having a git repo

Wouldn't it be better if that precondition would only run if we detect a valid git repo in the first place? Then we wouldn't need an option to disable that.

manuel3108 (approved):
AFAIK there is only that one thing remaining regarding the types for option conditions. Everything else seems to be resolved and working as expected.

I would suggest merging and publishing this, everything else could happen in other PR in my opinion. Any thoughts?

AdrianGonz97:
> I would suggest merging and publishing this, everything else could happen in other PR in my opinion. Any thoughts?

Just needed to add back the condition that checks for the command for publishing previews. Besides the issue with types for option conditions, I'm cool with merging this too 🙂

packages/core/adder/options.ts
manuel3108: See #397

manuel3108:
Thanks @AdrianGonz97 for the implementation.

This is now published: `npx @svelte-add/drizzle --path ./temp`

Issue #367, opened 2024-05-19
docs: add documentation for creating community adders


manuel3108:
### Topic

docs

### Description

There is some minor information on how to create a new adder here: https://github.com/svelte-add/svelte-add/blob/main/CONTRIBUTING.md#create-new-adder

And a community template here: https://github.com/svelte-add/adder-template

But besides that there is no documentation.

### Describe the solution you'd like

Decide on where the docs should live
* website (don't think so)
* GH wiki
* inside git repo

PR #382, opened 2024-06-06
chore: Convert `svelte-add.com` to TS

AdrianGonz97:
Should probably be merged after #379


benmccann:
There's an effort to overhaul svelte.dev and integrate the Svelte Add documentation into it to make it more visible and officially supported, so I don't know if you want to spend too much time on `svelte-add.com`

manuel3108 (approved):

manuel3108:
Will be merged after #379 

@benmccann Sound great! Maybe we can have a chat about that, also in regards of #367 (and where it should live)

Issue #358, opened 2024-05-12
storybook tests are failing


manuel3108:
### Topic

tests, storybook

### Version

invalid

### Expected behavior

Test's should be passing.

### Actual behavior

As found out by #357, storybook tests are currently failing. This is due to a misconfiguration in the `storybook` cli that fails to apply the changes to the `package.json` in certain circumstances. This issue has been reported to the remote and is currently tracked by https://github.com/storybookjs/storybook/issues/27096

As described in the remote issue, removing `--skip-install` fixes this behavior, but will therefore prompt the user to install the dependencies twice, once by the storybook cli, and once by `svelte-add`. This could be solved within `svelte-add`, although I would prefer that the storybook cli works correctly.

PR #384, opened 2024-06-11
chore: use `pnpm` for tests

AdrianGonz97:
This PR (which was originally part of #379, but is now separated into it's own) adjusts the testing suite's execs to use `pnpm` instead of `npm`, just like in #362, as it's leagues faster during the dependency install phase (`npm` would often hang and slow to a crawl when testing locally). 

Unlike #362, where some tests were failing in CI, I've tweaked the workflow to use pnpm's recommended setup with their action (https://github.com/pnpm/action-setup), which fixes the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules/
 .pnpm-store/
 build/
+.npmrc
 packages/tests/.outputs
 temp/
 coverage/


.github/workflows/changesets.yml
```diff
@@ -3,9 +3,6 @@ on:
     push:
         branches:
             - main
-env:
-    CI: true
-    PNPM_CACHE_FOLDER: .pnpm-store
```
AdrianGonz97, line 8: Removed these as `PNPM_CACHE_FOLDER` is no longer necessary and `CI` is already an applied env var (see: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

package.json
```diff
@@ -19,7 +19,8 @@
         "utils:readmes": "node ./packages/dev-utils/build/index.js readmes",
         "utils:packages": "node ./packages/dev-utils/build/index.js packages",
         "utils:dependencies": "node ./packages/dev-utils/build/index.js dependencies && pnpm prettier:fix",
-        "check": "pnpm eslint:check && pnpm prettier:check && pnpm types:check"
+        "check": "pnpm eslint:check && pnpm prettier:check && pnpm types:check",
+        "postinstall": "node packages/testing-library/node_modules/puppeteer/install.mjs"
```
AdrianGonz97, line 23: This `postinstall` script should already exist in puppeteer's `package.json`. For some unknown reason, it's not executing in CI for us. Applying it here manually fixes that.

package.json
```diff
@@ -46,5 +47,6 @@
         "rollup-plugin-add-shebang": "^0.3.1",
         "rollup-plugin-preserve-shebangs": "^0.2.0",
         "typescript": "^5.4.5"
-    }
+    },
+    "packageManager": "pnpm@9.3.0"
```
AdrianGonz97, line 51: Added this field as the pnpm action will auto-install the version that's specified here (see: https://github.com/svelte-add/svelte-add/pull/379#discussion_r1633657080).

package.json
benmccann: I wonder if it's worth trying to use playwright instead since that's what we use the other SvelteJS projects? It looks like a relatively small number of places that puppeteer is referenced and I know we don't have to do this with Playwright. Though we do some custom stuff on the CI which I believe is to only download the relevant browser instead of all of them to speed things up: https://github.com/sveltejs/kit/blob/main/.github/workflows/ci.yml

Anyway, if we want to stick with puppeteer, I wonder if we should try setting the side-effects -cache to false instead as I think that's probably what's happening: https://pnpm.io/npmrc#side-effects-cache

packages/testing-library/utils/workspace.ts
```diff
@@ -11,7 +11,7 @@ export function getTemplatesDirectory(options: TestOptions) {
 
 export async function installDependencies(output: string) {
     try {
-        await executeCli("npm", ["install"], output, { stdio: "pipe" });
+        await executeCli("pnpm", ["install", "--ignore-workspace"], output, { stdio: "pipe" });
```
benmccann, line 14: it could be worth a comment here as to why we need `--ignore-workspace`

.github/workflows/changesets.yml
```diff
@@ -21,19 +18,30 @@ jobs:
               with:
                   fetch-depth: 0
 
+            - name: setup pnpm
+              uses: pnpm/action-setup@v4
+
             - name: setup node.js
               uses: actions/setup-node@v3
```
benmccann, line 25: could you bump `checkout` and `setup-node` to v4 as long as we're updating this?

.github/workflows/changesets.yml
```diff
@@ -21,19 +18,30 @@ jobs:
...
               with:
                   node-version: 20
```
benmccann, line 27: you can also add a line here `cache: pnpm` which will use a pre-installed version of pnpm to save you from having to download and install it

.github/workflows/changesets.yml
```diff
@@ -21,19 +18,30 @@ jobs:
...
-            - name: install pnpm
-              run: npm i pnpm@latest -g
+
+            # PNPM Store cache setup
+            - name: get pnpm store directory
+              id: pnpm-cache
+              run: |
+                  echo "pnpm_cache_dir=$(pnpm store path)" >> $GITHUB_OUTPUT
```
benmccann, line 33: do we need this part? I'm not quite sure what it does, but it's not present in pnpm's docs (https://github.com/pnpm/action-setup) and it seems like it'd save a lot of lines to remove it wherever it shows up
```suggestion
            - name: get pnpm store directory
              id: pnpm-cache
              run: |
                  echo "pnpm_cache_dir=$(pnpm store path)" >> $GITHUB_OUTPUT
```

AdrianGonz97: It's there, at the bottom of the README here: https://github.com/pnpm/action-setup#use-cache-to-reduce-installation-time

In the past, I've found using https://github.com/actions/cache directly to be far more reliable than using the built-in `cache` option in https://github.com/actions/setup-node. Perhaps it's better nowadays? I'd be happy to give it another go

benmccann: oh, their setup is a bit different. e.g. they use `GITHUB_ENV` instead of `GITHUB_OUTPUT`, which is why I didn't find it when I did "Ctrl+F" for "GITHUB_OUTPUT"

anyway, we've been using the built-in `cache: 'pnpm'` in SvelteKit for ages and I haven't noticed any issues with it, but I don't know what issues you've seen

AdrianGonz97: > but I don't know what issues you've seen

back when i was setting it up for skeleton (probably a little over a year ago now) it wasn't actually caching for some reason 🤷

it looks like it's working here though!

benmccann:
Hmm. I wonder what the test failure regarding `storybook` is about. The CI seems to be passing on `main`

AdrianGonz97:
> Hmm. I wonder what the test failure regarding `storybook` is about. 

@manuel3108 raised an issue for it here: https://github.com/svelte-add/svelte-add/issues/358

it could be solved by removing the `--skip-install` flag, though, it may have some undesirable results that can conflict with our install flow

> The CI seems to be passing on `main`

As of now, the `pr.yml` workflow is only ever ran on PRs, so that's why the failing workflow isn't reflected on `main`

package.json
AdrianGonz97: > I wonder if it's worth trying to use playwright instead since that's what we use the other SvelteJS projects? It looks like a relatively small number of places that puppeteer is referenced and I know we don't have to do this with Playwright.

I tried doing this a couple of weeks ago _before_ I had a good understanding of the codebase and failed magnificently 😄 Wasn't as clean of a swap-out that I thought it would be. I'm willing to give it another shot though

But yea, regarding tests, I'd really like to spend some time improving the testing suite as a whole. atm, it's a bit slow and the tests aren't truly executing in parallel (despite the use of the, deceptively named, [`promise-parallel-throttle`](https://github.com/svelte-add/svelte-add/blob/62db0c9be7f737dcd850968d6b936e671a84c935/packages/testing-library/package.json#L8) package)

> Anyway, if we want to stick with puppeteer, I wonder if we should try setting the side-effects -cache to false instead as I think that's probably what's happening: [pnpm.io/npmrc#side-effects-cache](https://pnpm.io/npmrc#side-effects-cache)

thanks for the suggestion! I'll see if this works

benmccann:
> As of now, the pr.yml workflow (which contains the tests job) is only ever ran on PRs, so that's why the failing workflow isn't reflected on main

Oh, I see. Perhaps we could update that? We could use the following:
```
on:
  push:
    branches:
      - main
  pull_request:
```

benmccann (approved):
nice!! so many great updates here. thank you so much for all of this

I pinged the Storybook folks to see if they might be able to look at that - though I think the guy I've been working with most lately might be on paternity leave at the moment

.npmrc
```diff
@@ -0,0 +1 @@
+side-effects-cache=false
```
benmccann, line 1: we should probably merge https://github.com/svelte-add/svelte-add/pull/385 first and see if it allows us to remove this file

manuel3108 (approved):

.npmrc
manuel3108: After merging #385 and removing this file the pipeline is still working. Thanks for this note.

PR #385, opened 2024-06-11
chore: replace `puppeteer` for `playwright`

AdrianGonz97:
With a little bit of finagling, the migration to playwright was relatively straightforward (compared to my last attempt, where I was trying to also integrate vitest as well). 

I'll be leaving some comments attached for reasons behind specific changes.

### DIFF
--- a/adders/bootstrap/config/tests.js
+++ b/adders/bootstrap/config/tests.js
@@ -51,7 +51,7 @@ export const tests = defineAdderTests({
         {
             name: "typography properties",
             run: async ({ expectProperty }) => {
-                await expectProperty(".h1", "font-size", "34px");
+                await expectProperty(".h1", "font-size", "40px");
                 await expectProperty(".text-decoration-line-through", "text-decoration", "line-through solid rgb(33, 37, 41)");
             },
         },
--- a/adders/bootstrap/config/tests.js
+++ b/adders/bootstrap/config/tests.js
@@ -58,7 +58,7 @@ export const tests = defineAdderTests({
         {
             name: "dropdown properties",
             run: async ({ click, elementExists }) => {
-                await click(".btn.dropdown-toggle", false);
+                await click(".btn.dropdown-toggle");
                 await elementExists(".btn.dropdown-toggle.show");
             },
         },
--- a/adders/routify/config/tests.js
+++ b/adders/routify/config/tests.js
@@ -12,10 +12,10 @@ export const tests = defineAdderTests({
                 await elementExists(".routify-demo");
                 await expectUrlPath("/");
 
-                await click(".routify-demo .demo", true);
+                await click(".routify-demo .demo", "/demo");
                 await expectUrlPath("/demo");
 
-                await click(".routify-demo .index", true);
+                await click(".routify-demo .index", "/");
                 await expectUrlPath("/");
             },
         },
--- a/packages/testing-library/utils/test.ts
+++ b/packages/testing-library/utils/test.ts
@@ -1,6 +1,6 @@
-import { AdderWithoutExplicitArgs, Tests } from "@svelte-add/core/adder/config";
-import { OptionValues, Question } from "@svelte-add/core/adder/options";
-import { Page } from "puppeteer";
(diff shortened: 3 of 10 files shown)


adders/bootstrap/config/tests.js
```diff
@@ -51,14 +51,14 @@ export const tests = defineAdderTests({
         {
             name: "typography properties",
             run: async ({ expectProperty }) => {
-                await expectProperty(".h1", "font-size", "34px");
+                await expectProperty(".h1", "font-size", "40px");
```
AdrianGonz97, line 54: the default viewport size of puppeteer is 800x600, while playwright's is 1280x720. given that `.h1` uses `rem` to for the font-size, this value needed to be tweaked. 

thankfully it was the only test that needed adjusting in this regard

packages/testing-library/utils/test.ts
```diff
@@ -44,15 +44,21 @@ async function elementExists(page: Page, selector: string) {
...
-    if (!waitForNavigation) {
-        await page.click(selector);
-    } else {
-        // if a click triggers a page reload, this is the correct
-        // syntax according to puppeteer documentation
-        await Promise.all([page.waitForNavigation(), page.click(selector)]);
+    await page.click(selector);
+
+    if (path) {
+        await page.waitForURL((url) => url.pathname === path);
```
AdrianGonz97, line 61: while `page.waitForNavigation()` _is_ an available api in playwright, it's deprecated. instead, it recommends to use `page.waitForURL`, which requires validation as to what we expect the url path to be, hence the change of `waitForNavigation` to `path`

adders/routify/config/tests.js
```diff
@@ -12,10 +12,10 @@ export const tests = defineAdderTests({
                 await elementExists(".routify-demo");
                 await expectUrlPath("/");
 
-                await click(".routify-demo .demo", true);
+                await click(".routify-demo .demo", "/demo");
                 await expectUrlPath("/demo");
 
-                await click(".routify-demo .index", true);
+                await click(".routify-demo .index", "/");
                 await expectUrlPath("/");
```
AdrianGonz97, line 19: I wasn't 100% sure about this, but perhaps we can combine these cases? e.g. if a `path` is passed to `click`, then `click` will run `expectUrlPath` for you since `click` already knows the expected path

benmccann (approved):
so nice!!

AdrianGonz97:
hmm, have a little bit of a conundrum here. it looks like CI is [failing here](https://github.com/svelte-add/svelte-add/actions/runs/9474158135/job/26103197218?pr=385) because i forgot to install the browsers, but I'm also trying to avoid touching the workflows here to prevents conflicts with the other PR 😅

i think we can temporarily solve both by adding the `postinstall` script back in here, like: `"postinstall": "playwright install chromium"`, and then adjust the workflow files in the other PR accordingly after this PR has merged

what do you think @benmccann?

benmccann:
Sounds fine to me. Thanks again for doing this!!

adders/routify/config/tests.js
manuel3108: Yes, I think we could, especially now that we need to pass the path a second time. But I think the current way is more explicit, and does not produce unwanted side effects.

manuel3108 (approved):

manuel3108:
@AdrianGonz97 As for your comment regarding `vitest`, that's what I initially tried to implement after seeing that uvu has [some serious problems with multithreading](https://github.com/lukeed/uvu/issues/14)

But since `vitest` is bundeling everything on it's own, I had serious problems with all the shebangs required in the clis. That's why i decided to leave it as is and implement some custom logic for testing.

manuel3108:
> i think we can temporarily solve both by adding the `postinstall` script back in here, like: `"postinstall": "playwright install chromium"`, and then adjust the workflow files in the other PR accordingly after this PR has merged

@AdrianGonz97 Both PRs are now merged, you can now proceed with this change. Keep in mind that if someone checks out the repo for the first time, it test's should still be working for that person. Same goes for consumers of `@svelte-add/testing-library` that can be consumed separately, i.e. as part of a community adder ([template](https://github.com/svelte-add/adder-template))

Issue #339, opened 2024-05-05
Github Pages


manuel3108:
### Website

a

### Documentation

a

### Integration summary

1. install package `tool`
2. modify `svelte-config.js` and to add tool as preprocessor
3. import tool inside `+layout.svelte`
4. and so on...

AdrianGonz97:
I think this would also likely be out of the scope of being part of the core of svelte-add. Could be a good community adder though

PR #388, opened 2024-06-13
feat: post-conditions

### LINKED ISSUE #337: Add pre / post-installation checks
### Topic

core, tests

### Description

Currently it is possible to apply the same composer multiple times to a project. Additionally, we don't have a way to detect if a composer is supported on the current computer (external dependencies installed, see `tauri` which requires `rust`)

I have previously already added some types to get a feeling for all components of a composer. The types are not accurate and do not provide any meaningful enhancements. Everything needs to be re-thinked.

#### Preconditions
Preconditions need to ensure that a given composer can be used within the given environment. This is especially useful, when considering composers that require some other globally installed tooling to work correctly (#9). Preconditions are workspace independent and can only check for the global environment.

Here is a sample pre installation check to determine if the Tauri adder could be installed:
```js
export const checks = defineComposerChecks({
    options,
    preconditions: [
        {
            name: "rust installed",
            run: ({ execute }) => {
                // execute a shell command and gather it's output and exit code
                const { output, exitCode } = execute("cargo", ["version"]);

                // if the exit code is 0, we can assume that cargo / rust is installed correctly
                return exitCode == 0;
            },
        },
    ],
    preInstallation: [],
    postInstallation: [],
});
```

Precondition checks should be executed before the composer runs to avoid unnecessary file modifications in case any of the preconditions fails. In case at least one precondition fails, the composers should not install itself and rather fail. The user should be notified with an appropriate error message.

#### Pre installation checks
Pre installation checks are executed before the adder executes, but after the project has been detected / created and therefore have full access to the workspace (minus the options, as they have not been chosen yet). These checks should identify if 
- the composer is supported in the current environment (svelte / kit)
- the composer has already been applied or the tool to compose is already setup (i.e. the `tailwdincss.config.js` file is already present / installed packages already present)

#### Post installation checks
Should check if a composer was successfully installed, basically they are the same as the pre installation checks, but they are executed after a composer was executed to determine if everything was set up as expected.

#### Further thoughts
Those ideas are mainly taken from `svelte-add` (predecessor of `svelte-compose`) and need to be partially re-thought. 

**Do we really need the post installation checks?**
Is there any reason we would not trust in our code? Additionally we have tests that run when merging PR's that ensure that alle test execute correctly. Those tests not only check that that special files are present and contain special contents, but run explicit integration tests for each possible composer and svelte template possibility. 

**Do we really need pre installation checks?**
Once we got this far, let's think if we need pre installation checks. Should they be optional? For most composers it should be enough to check if any of the dependencies is already present. But what if we have something like #18 (Github pages) that does not apply a dependency? Should we provide pre build checks for checking the dependencies? How should the method definition of those look like?

manuel3108:
Closes #337 (preconditions already implemented by #368) 

Usage:
```js
export const checks = defineAdderChecks({
    options,
    postconditions: [
        {
            name: "scss setup",
            run: async ({ fileExists, fileContains }) => {
                if (options.useSass) {
                    const filePath = `src/variables.scss`;
                    await fileExists(filePath);
                    await fileContains(filePath, "$background: lightgrey;");
                }
            },
        },
    ],
});
```

Will be helpful for all adders that we cannot check with integration testing, like #338, #339 , etc.

This is currently **not** implemented for any adder, as all adders are currently testable by integration testing. Future adders will use this feature.

Console output looks like this:
image

### DIFF
--- a/.changeset/mighty-knives-count.md
+++ b/.changeset/mighty-knives-count.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/core": minor
+---
+
+add postconditions
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,4 @@
 {
     "typescript.tsserver.experimental.enableProjectDiagnostics": true,
-    "cSpell.words": ["ecma", "popperjs", "Svex", "threlte"]
+    "cSpell.words": ["ecma", "popperjs", "postcondition", "Postcondition", "postconditions", "Svex", "threlte"]
 }
--- a/adders/bootstrap/config/checks.js
+++ b/adders/bootstrap/config/checks.js
@@ -9,7 +9,7 @@ export const checks = defineAdderChecks({
             run: async ({ fileExists, fileContains }) => {
                 const filePath = `src/variables.scss`;
                 await fileExists(filePath);
-                await fileContains(filePath, "testasd");
+                await fileContains(filePath, "$background: lightgrey;");
             },
         },
     ],
--- a/adders/bootstrap/config/checks.js
+++ b/adders/bootstrap/config/checks.js
@@ -3,4 +3,14 @@ import { options } from "./options";
 
 export const checks = defineAdderChecks({
     options,
+    postconditions: [
+        {
+            name: "scss setup",
+            run: async ({ fileExists, fileContains }) => {
+                const filePath = `src/variables.scss`;
+                await fileExists(filePath);
+                await fileContains(filePath, "testasd");
+            },
+        },
+    ],
 });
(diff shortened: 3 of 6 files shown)


manuel3108:
@AdrianGonz97 Would you mind reviewing this, even if this is not directly testable?

AdrianGonz97 (commented):
Code-wise everything looks good! although, I do have a few questions:

So if a postcondition fails for a user, what are the actionable steps that a user should take when an error is thrown? Should it tell the user how to fix it (_if it can be fixed_)? Should it provide a link to our issue tracker and ask them to raise an issue?

manuel3108:
Good catch. I added an appropriate text to inform the user that they should create an issue.

image

PR #392, opened 2024-06-13
chore: bump dependencies

manuel3108:


adders/tailwindcss/config/adder.js
```diff
@@ -18,13 +18,13 @@ export const adder = defineAdderConfig({
...
     packages: [
-        { name: "tailwindcss", version: "^3.4.3", dev: true },
+        { name: "tailwindcss", version: "^3.4.4", dev: true },
         { name: "postcss", version: "^8.4.38", dev: true },
         { name: "autoprefixer", version: "^10.4.19", dev: true },
         { name: "@tailwindcss/typography", version: "^0.5.13", dev: true, condition: ({ options }) => options.typography },
         {
             name: "prettier-plugin-tailwindcss",
-            version: "^0.5.14",
+            version: "^0.6.4",
```
benmccann, line 27: this is the only one that actually has an effect. the user gets the latest version for all the rest anyways because of semver, so next time you could probably avoid needing to bump all the various packages

manuel3108: Ohh, I did not know that. I will update the script to take care of that!

Issue #386, opened 2024-06-12
chore: Revert storybook adder back to `@latest` when `v8.3.0` releases


AdrianGonz97:
### Topic

storybook

### Description

As of #384, we've changed the storybook version to use the `@next` tag as it fixed a the bug described in #358. When `8.2.0` releases, we should revert the adder back to using `@latest`.

### Describe the solution you'd like

- [x] Swap `@next` tag for `@latest`
- [x] Remove `--legacy-peer-deps` flag from `npm install`: https://github.com/svelte-add/svelte-add/commit/668c6c797b27b09b752841aaebdeb8c8a58af2f4

benmccann:
We should also remember to remove the `--legacy-peer-deps` flag: https://github.com/svelte-add/svelte-add/commit/668c6c797b27b09b752841aaebdeb8c8a58af2f4

benmccann:
They said the fix is in 8.1.8, so we should be able to try `latest` again

manuel3108:
@benmccann This is apparently not the case, or it's still not working as expected.
See #393, or better my [small repro workflow run](https://github.com/manuel3108/storybook-missing-packages-repro/actions/runs/9503757087/job/26194895861), which clearly shows that 8.1.8 is used, but without any changes in the output.

Interestingly enough that PR [appeared in the Changelog for 8.1.8](https://github.com/storybookjs/storybook/blob/next/CHANGELOG.md), but seems to be [re-included for the changelog for 8.1.9](https://github.com/storybookjs/storybook/pull/28231)

manuel3108:
See #393 

In summary, `8.2.5` fixes some issues, but introduces other. As the currently published way is broken, let's still use `next` for now. Leaving this issue open for the same reason.

PR #393, opened 2024-06-13
chore: Revert storybook adder back to @latest when v8.2.0 releases

manuel3108:
Relates #386 

- Revert "add `--legacy-peer-deps` flag for npm installs"
- chore: Revert storybook adder back to @latest when v8.2.0 releases


benmccann:
Ah, I guess this doesn't work yet because it doesn't have the other fix that's in 8.2.0. But 8.2.0 should be released in the next week anyway, so I guess we can just wait a bit. Thanks for checking!

benmccann:
Storybook 8.2 is finally out! The latest is currently 8.2.1: https://www.npmjs.com/package/storybook

manuel3108:
They also fixed https://github.com/storybookjs/storybook/issues/28474 and https://github.com/storybookjs/storybook/issues/28475 which were happening on the latest 8.2 beta build.

But they managed to re-introduce https://github.com/storybookjs/storybook/issues/27096, which again disappears when using `next` ([`8.3.0-alpha.0`](https://www.npmjs.com/package/storybook/v/8.3.0-alpha.0)). They must have moved some commits around. 

Should we release this with the beta again, or should we report this issue again and wait for it to be resolved?

manuel3108:
Just had a look locally, and 8.2.5 is still not working as it should. The currently published implementation is broken, so let's use the next version to make it much better already. 

Let's leave the original issue open though, to provide further tracking.

PR #395, opened 2024-06-14
fix: move storybook from styling category to tools category

benmccann:
Storybook feels out-of-place in the styling category to me since it's not a CSS framework like the others and does so much more than just help you with styling. 

From their homepage:
Storybook is a frontend workshop for building UI components and pages in isolation. Thousands of teams use it for UI development, testing, and documentation.


manuel3108:
Good catch, don't know how it even ended there in the first place.

PR #401, opened 2024-06-14
fix: only print failed preconditions

### LINKED ISSUE #396: preconditions: only print failed preconditions
> We probably don't need to print the preconditions if they passed

_Originally posted by @benmccann in https://github.com/svelte-add/svelte-add/issues/379#issuecomment-2166905263_
            

Context:
Currently we print all preconditions (failed and passed). Similar to the recently implement post-conditions, we should only print failed preconditions, and only print the appropriate message, if at least one precondition failed.

manuel3108:
Closes #396

### DIFF
--- a/packages/core/adder/preconditions.ts
+++ b/packages/core/adder/preconditions.ts
@@ -101,8 +101,8 @@ export async function validatePreconditions<Args extends OptionDefinition>(
 
     if (allPreconditionsPassed) {
         return;
-    } 
-    
+    }
+
     if (isTesting) {
         throw new Error(`Preconditions failed: ${preconditionLog.join(" / ")}`);
     }
--- a/packages/core/adder/preconditions.ts
+++ b/packages/core/adder/preconditions.ts
@@ -101,20 +101,17 @@ export async function validatePreconditions<Args extends OptionDefinition>(
 
     if (allPreconditionsPassed) {
         return;
-    } else if (!allPreconditionsPassed && isTesting) {
+    } 
+    
+    if (isTesting) {
         throw new Error(`Preconditions failed: ${preconditionLog.join(" / ")}`);
     }
 
-    let allMessages = "";
-    for (const [i, message] of preconditionLog.entries()) {
-        allMessages += `- ${message}${i == preconditionLog.length - 1 ? "" : "\n"}`;
-    }
+    const allMessages = preconditionLog.map((msg) => `- ${msg}`).join("\n");
 
     messagePrompt("Preconditions not met", allMessages);
 
-    if (!allPreconditionsPassed) {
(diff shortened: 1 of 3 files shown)


packages/core/adder/preconditions.ts
```diff
@@ -86,32 +86,34 @@ export async function validatePreconditions<Args extends OptionDefinition>(
...
 
-        if (!allPreconditionsPassed && isTesting) {
-            throw new Error(`Preconditions failed: ${preconditionLog.join(" / ")}`);
-        }
-        if (isTesting) return;
+    if (allPreconditionsPassed) {
+        return;
+    } else if (!allPreconditionsPassed && isTesting) {
+        throw new Error(`Preconditions failed: ${preconditionLog.join(" / ")}`);
+    }
```
AdrianGonz97, line 106: since we're returning early when they all pass, we don't really need to check if it's still `false`
```suggestion
    if (allPreconditionsPassed) {
        return;
    } 
    
    if (isTesting) {
        throw new Error(`Preconditions failed: ${preconditionLog.join(" / ")}`);
    }
```

packages/core/adder/preconditions.ts
```diff
@@ -86,32 +86,34 @@ export async function validatePreconditions<Args extends OptionDefinition>(
...
+
+    let allMessages = "";
+    for (const [i, message] of preconditionLog.entries()) {
+        allMessages += `- ${message}${i == preconditionLog.length - 1 ? "" : "\n"}`;
+    }
```
AdrianGonz97, line 111: I find this to be a bit easier to parse, though it may just be a preference:
```suggestion
    const allMessages = preconditionLog.map((msg) => `- ${msg}`).join("\n");
```

packages/core/adder/preconditions.ts
```diff
@@ -86,32 +86,34 @@ export async function validatePreconditions<Args extends OptionDefinition>(
...
 
-        messagePrompt("Preconditions:", allMessages);
+    messagePrompt("Preconditions not met", allMessages);
 
-        if (!allPreconditionsPassed) {
-            await askUserToContinueWithFailedPreconditions();
-        }
+    if (!allPreconditionsPassed) {
+        await askUserToContinueWithFailedPreconditions();
     }
```
AdrianGonz97, line 117: same as the first comment - since we're returning early if they all pass, we can safely remove this condition
```suggestion
    await askUserToContinueWithFailedPreconditions();
```

AdrianGonz97 (changes requested):
looks good! just some have some minor nits:

packages/core/adder/preconditions.ts
manuel3108: Don't know what my brain was thinking there, you are 150% right!

PR #402, opened 2024-06-14
fix: avoid adding adders to unsupported project types

### LINKED ISSUE #387: preconditions: evaluate valid environments
### Topic

core

### Version

@svelte-add/drizzle@1.0.0

### Expected behavior

The adder is asking to create a new project. I select to create a `svelte` project (non `kit`). The adder continues applying itself without warnings, even tho it specifies that it should only be working on kit.

Same thing can happen on existing projects.

### Actual behavior

I think we can re-use preconditions for that. We already have a global precondition for dirty git workspaces. Let's add a global precondition that fails, when you are trying to setup at least one adder for an invalid project.

### LINKED ISSUE #394: Don't offer to add Routify to SvelteKit projects
### Topic

routify

### Version

2.0.10

### Expected behavior

It doesn't really make sense to setup routify on a SvelteKit project, so that one should just be left out of the list of options. We could probably check if `@sveltejs/kit` is in either `dependencies` or `devDependencies`

I almost wonder if we should just remove the Routify adder. [SvelteKit has 50x more usage](https://npmcharts.com/compare/@roxi/routify,@sveltejs/kit?log=false&interval=30) and is the officially recommended solution. With svelte-add becoming more official it feels a little weird not to offer an option for the one we recommend folks use and offer an option for another one

### Actual behavior

When I run `npx svelte-add`, it offers to add mdsvex, routify, and Threlte as tools

manuel3108:
Overall, this PR tries to avoid that users get the possibility to apply an adder for an unsupported environment. Each adder can define the supported environments in its adder config.

Closes #394 
Closes #387

There are a few cases that we need to consider in order to ensure this requirement.

```sh
npx svelte-add
```
This will detect a project in the cwd or ask the user to create a project. After that it will display all categories with the appropriate adders. Up to this point a project has been detected for sure, so we can just hide the invalid combinations (i.e. routify, if kit project)

```sh
npx @svelte-add/routify
```
This will detect a project in the cwd or ask the user to create a project. If creating a project, the user is unable to create a kit project, as that option is hidden, and the svelte project template is automatically selected.
In case this was called on an existing project there is a new precondition that will fail, indicating that routify is not supported in kit projects.

```sh
npx svelte-add --adder bulma routify
```
This will detect a project in the cwd or ask the user to create a project. If creating a project, we cannot hide either the kit or svelte templates as both templates are valid for at least one adder. Both options remain available for the user.
In case this was called on an existing project, or on a new project there is a new precondition that will fail, indicating that routify is not supported in kit projects.

---

I hope all of this kinda makes sense. I tried to find a more unified approach but all of these approaches had problems with the user flow. This is the best, un-obtrusive solution I came up with.

But I'm happy to implement a better solution is someone can come up with one.

### DIFF
--- a/packages/core/adder/execute.ts
+++ b/packages/core/adder/execute.ts
@@ -1,6 +1,6 @@
 import path from "path";
 import { commonFilePaths, format, writeFile } from "../files/utils.js";
-import { ProjectType, createProject, detectSvelteDirectory } from "../utils/create-project.js";
+import { type ProjectType, createProject, detectSvelteDirectory } from "../utils/create-project.js";
 import { createOrUpdateFiles } from "../files/processors.js";
 import { Package, executeCli, getPackageJson, groupBy } from "../utils/common.js";
 import {
--- a/packages/core/adder/execute.ts
+++ b/packages/core/adder/execute.ts
@@ -1,6 +1,6 @@
 import path from "path";
 import { commonFilePaths, format, writeFile } from "../files/utils.js";
-import { createProject, detectSvelteDirectory } from "../utils/create-project.js";
+import { ProjectType, createProject, detectSvelteDirectory } from "../utils/create-project.js";
 import { createOrUpdateFiles } from "../files/processors.js";
 import { Package, executeCli, getPackageJson, groupBy } from "../utils/common.js";
 import {
@@ -98,6 +98,7 @@ async function executePlan<Args extends OptionDefinition>(
 ) {
     const remoteControlled = remoteControlOptions !== undefined;
     const isTesting = remoteControlled && remoteControlOptions.isTesting;
+    const isExecutingMultipleAdders = adderDetails.length > 1;
 
     if (!isTesting) startPrompts(`Welcome to ${executingAdder.name}@${executingAdder.version}`);
 
@@ -109,13 +110,17 @@ async function executePlan<Args extends OptionDefinition>(
         executionPlan.workingDirectory = directory;
     }
 
+    const workspace = createEmptyWorkspace();
+    await populateWorkspaceDetails(workspace, executionPlan.workingDirectory);
+    const projectType: ProjectType = workspace.kit.installed ? "kit" : "svelte";
+
     // select appropriate adders
     let userSelectedAdders = executionPlan.commonCliOptions.adders ?? [];
-    if (userSelectedAdders.length == 0 && adderDetails.length > 1) {
+    if (userSelectedAdders.length == 0 && isExecutingMultipleAdders) {
         // if the user has not selected any adders via the cli and we are currently executing for more than one adder
         // the user should have the possibility to select the adders he want's to add.
-        userSelectedAdders = await askForAddersToApply(adderDetails);
-    } else if (userSelectedAdders.length == 0 && adderDetails.length == 1) {
+        userSelectedAdders = await askForAddersToApply(adderDetails, projectType);
+    } else if (userSelectedAdders.length == 0 && !isExecutingMultipleAdders) {
         // if we are executing only one adder, then we can safely assume that this adder should be added
         userSelectedAdders = [adderDetails[0].config.metadata.id];
     }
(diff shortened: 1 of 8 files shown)


packages/core/utils/create-project.ts
```diff
@@ -1,28 +1,11 @@
 import * as path from "path";
-import { booleanPrompt, selectPrompt, textPrompt } from "./prompts.js";
+import { PromptOption, booleanPrompt, selectPrompt, textPrompt } from "./prompts.js";
 import { commonFilePaths, directoryExists, fileExists } from "../files/utils.js";
```
AdrianGonz97, line 3: we'll want to denote `PromptOption` as a type and add `endPrompts` (see below)
```suggestion
import { type PromptOption, booleanPrompt, selectPrompt, textPrompt, endPrompts } from "./prompts.js";
```

packages/core/utils/create-project.ts
```diff
@@ -56,7 +39,7 @@ export async function detectSvelteDirectory(directoryPath: string): Promise<stri
     return null;
 }
 
-export async function createProject(cwd: string) {
+export async function createProject(cwd: string, supportKit: boolean, supportSvelte: boolean) {
     const createNewProject = await booleanPrompt("Create new Project?", true);
     if (!createNewProject) {
         console.log("New project should not be created. Exiting.");
```
AdrianGonz97, line 45: this should be using `endPrompts` rather than `console.log` - it's also reiterating what's already said in the prompt, so we can just remove it and set it to just `Exiting.`

```suggestion
        endPrompts("Exiting.");
```

packages/core/adder/preconditions.ts
```diff
@@ -40,6 +43,29 @@ function getGlobalPreconditions(
                     }
                 },
             },
+            {
+                name: "supported environments",
+                // eslint-disable-next-line @typescript-eslint/require-await
```
AdrianGonz97, line 48: we can get rid of the need for this line by tweaking the type definition of `run`. rather than using a `Promise`, we can use a `MaybePromise`:
```ts
type MaybePromise<T> = Promise<T> | T;

// core/adder/config.ts
 export type Precondition = {
    name: string;
    run: () => MaybePromise<{ success: boolean; message: string | undefined }>;
};
```
we wouldn't have to change anything else either, since `await` at the invocation sites can work for both an `async` and regular function

packages/core/adder/preconditions.ts
```diff
@@ -40,6 +43,29 @@ function getGlobalPreconditions(
...
+                    });
+
+                    if (addersForInvalidEnvironment.length == 0) {
+                        return { success: true, message: undefined };
+                    }
+
+                    const messages: string[] = [];
+                    for (const adder of addersForInvalidEnvironment) {
+                        messages.push(`"${adder.config.metadata.name}" does not support "${projectType}"`);
+                    }
```
AdrianGonz97, line 65: another nit, though it's a matter of preference:
```suggestion
                    const messages = addersForInvalidEnvironment.map(
                        (adder) => `"${adder.config.metadata.name}" does not support "${projectType}"`,
                    );
```

(not sure why the suggestion on github is so rough with it on the formatting 🤔 )

packages/core/adder/execute.ts
```diff
@@ -1,6 +1,6 @@
 import path from "path";
 import { commonFilePaths, format, writeFile } from "../files/utils.js";
-import { createProject, detectSvelteDirectory } from "../utils/create-project.js";
+import { ProjectType, createProject, detectSvelteDirectory } from "../utils/create-project.js";
```
AdrianGonz97, line 3:
```suggestion
import { type ProjectType, createProject, detectSvelteDirectory } from "../utils/create-project.js";
```

packages/core/utils/create-project.ts
```diff
@@ -73,10 +56,15 @@ export async function createProject(cwd: string) {
...
-        { label: "Svelte", value: "svelte" },
-    ]);
+    const availableProjectTypes: PromptOption<string>[] = [];
+    if (supportKit) availableProjectTypes.push({ label: "SvelteKit", value: "kit" });
+    if (supportSvelte) availableProjectTypes.push({ label: "Svelte", value: "svelte" });
+
+    let projectType: string;
+
+    if (availableProjectTypes.length == 0) throw new Error("Failed to identify possible project types");
+    else if (availableProjectTypes.length == 1) projectType = availableProjectTypes[0].value;
```
AdrianGonz97, line 66: _the most minor of nits:_ `else` isn't necessary
```suggestion
    if (availableProjectTypes.length == 1) projectType = availableProjectTypes[0].value;
```

AdrianGonz97 (changes requested):

packages/core/adder/preconditions.ts
manuel3108: JS is so fancy sometimes...
Good catch, done.

PR #403, opened 2024-06-14
chore: have `pkg.pr.new` publish when a PR is "ready for review"

AdrianGonz97:
this PR also removes `pkg-pr-new` as a dependency and instead executes it remotely with `pnpm dlx pkg-pr-new@0.0`, which pins it to the latest patch release

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -1,11 +1,11 @@
 name: Publish Preview Release
 on:
     pull_request:
-        types: [ready_for_review, synchronize, opened]
+        types: [ready_for_review, synchronize, opened, labeled]
 
 jobs:
     preview-release:
-        if: github.repository == 'svelte-add/svelte-add' && !github.event.pull_request.draft
+        if: github.repository == 'svelte-add/svelte-add' && (!github.event.pull_request.draft || github.event.pull_request.labels.name == 'publish:preview')
         timeout-minutes: 30
         runs-on: ubuntu-latest
         steps:
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -1,7 +1,7 @@
 name: Publish Preview Release
 on:
     pull_request:
-        types: [ready_for_review, synchronize]
+        types: [ready_for_review, synchronize, opened]
 
 jobs:
     preview-release:
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -1,11 +1,10 @@
 name: Publish Preview Release
 on:
     pull_request:
-        types: [ready_for_review, synchronize, opened, labeled]
 
 jobs:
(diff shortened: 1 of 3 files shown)


AdrianGonz97:
hmm, the repo root _should not_ be publishing - looking into it

benmccann:
Publishing on every commit sounds fine. Why not also do it for draft PRs though? You could be playing around with something that you're not ready for people to review yet, but still want to see what it looks like

AdrianGonz97:
> Why not also do it for draft PRs though? You could be playing around with something that you're not ready for people to review yet, but still want to see what it looks like

my thought process is that if you're the creator of the PR, that means that you're likely editing it locally and that you should already be able to execute the CLI locally to test it as well - `pkg-pr-new` is more useful for those that are reviewing the PR and wanting to test the changes without having to `git checkout` it, rather than it being the main testing ground for the creator of the PR

manuel3108:
Do you think we could an additional condition somewhere? Like the pull request having a specific label? Just as an addition to how it's implemented right now. 

Because as of now, every random person opening a PR would cause a publish to stackblitz. That could also solve the problem with draft pull requests.

AdrianGonz97:
I think the current implementation is the best of both worlds. Users have control of when it publishes, but it doesn't publish _every_ commit, unless they want it to by putting it in a ready state.

> Do you think we could an additional condition somewhere? Like the pull request having a specific label? Just as an addition to how it's implemented right now.

I totally can. We'll just need to create a new label for it. is `publish:preview` alright?

benmccann:
Honestly I prefer doing it without a label. E.g. if we use a label then non-maintainers have no way to trigger a build and would need to ping maintainers

It does feel like doing it on every commit is a bit wasteful, but storage is cheap, so maybe stackblitz doesn't care. We can revisit if they add quotas. And maybe they'll get comments working in the meantime, which wouldn't need to be limited just to maintainers.

manuel3108:
I mean, who cares. Let's try it without the label. We can always add it back later if we find it helpful.

AdrianGonz97:
fair enough - looking at their FAQ (which is a WIP): https://github.com/stackblitz-labs/pkg.pr.new/pull/91/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

seems like they don't retain them for long, which is a fair tradeoff

I'll remove the conditions (except for the repo specifier one) and have it publish on every commit

benmccann:
> my thought process is that if you're the creator of the PR, that means that you're likely editing it locally and that you should already be able to execute the CLI locally to test it as well

yeah. though I was thinking that if you're pushing it to github then you're wanting to share it with others. there might be things you need to fix or address before the code is ready to review, but you still want to show off what you have so far and get feedback. I think the same as avoiding the label, who cares if we publish more?

PR #407, opened 2024-06-14
chore: improve welcome message

### LINKED ISSUE #406: Suggestion for welcome message
### Topic

svelte-add

### Description

I noticed the way Rich setup `create-svelte` is like:

Screenshot from 2024-06-14 09-01-15

I kind of prefer how this hides away the version information and package name into some lighter colored text and then uses a nicer string for the main text

Use this same format like:
```
svelte-add version 2.0.11

┌  Welcome to Svelte Add!
```

manuel3108:
Closes #406

### DIFF
--- a/.changeset/clever-dragons-sort.md
+++ b/.changeset/clever-dragons-sort.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/core": patch
+---
+
+chore: improve welcome message
--- a/packages/core/adder/execute.ts
+++ b/packages/core/adder/execute.ts
@@ -26,6 +26,7 @@ import { validatePreconditions } from "./preconditions.js";
 import { PromptOption, endPrompts, multiSelectPrompt, startPrompts } from "../utils/prompts.js";
 import { CategoryKeys, categories } from "./categories.js";
 import { checkPostconditions, printUnmetPostconditions } from "./postconditions.js";
+import { gray } from "picocolors";
 
 export type AdderDetails<Args extends OptionDefinition> = {
     config: AdderConfig<Args>;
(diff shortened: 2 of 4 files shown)


manuel3108:
@AdrianGonz97 Due to our changed files recognition the bot now only published `@svelte-add/core`. If I see that correctly, this is kinda useless, as you cannot run the core package directly. Do you have any other ideas how we could properly handle that? Otherwise we should probably just publish everything, and then disable the comment with the new option they provided. Then we could either add that hint to the PR template, or do something ourselfs. What do you think?

AdrianGonz97:
> Do you have any other ideas how we could properly handle that?

I think I can adjust it to also publish fellow project packages if they are dependencies of a modified package. Looking into it now!

benmccann (approved):
nice!!

AdrianGonz97:
@manuel3108 the more i thought about it, the more it became unreasonable to implement 😅. would you mind replacing the workflow in this PR as well? 

```yml
name: Publish Preview Release
on:
    pull_request:

jobs:
    preview-release:
        if: github.repository == 'svelte-add/svelte-add'
        timeout-minutes: 30
        runs-on: ubuntu-latest
        steps:
            - name: checkout code repository
              uses: actions/checkout@v3

            - uses: pnpm/action-setup@v4

            - name: setup node.js
              uses: actions/setup-node@v3
              with:
                  node-version: 20
                  cache: pnpm

            - name: install dependencies
              run: pnpm install

            - name: build
              run: pnpm build:prod

            - name: publish preview
              run: |
                  pnpm dlx pkg-pr-new@0.0 publish --pnpm --comment=off adders/* packages/*
```

benmccann:
Maybe just publish all of them is something under `packages` changes? And if something under `adders` changes then just publish that one?

AdrianGonz97:
> Maybe just publish all of them is something under packages changes? And if something under adders changes then just publish that one?

that'd work, but only to a certain extent. there are packages that also depend on other packages inside of `packages/*`, so if one of their dependencies changes, they wouldn't be published either

#408 should solve this now

PR #408, opened 2024-06-14
chore: publish deps of modified packages

AdrianGonz97:
goal: this PR should allow modified packages, in addition to their dependents, to be published.

for example:
if `@svelte-add/core` is modified then all of its dependents (i.e. every adder) will be published.

note: I don't know how to write bash and, frankly, the syntax looks worse than regex (😅), so i wrote a simple node script instead. i also tried writing to `stdout` and piping the result to `pnpm dlx pkg-pr-new ...`, but it kept erroring out for some reason. for now, it all executes through the node script.

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -45,4 +45,4 @@ jobs:
               env:
                   CHANGED_DIRS: ${{ steps.changed-files.outputs.all_changed_files }}
               run: |
-                  node scripts/get-deps-to-publish.js | pnpm dlx pkg-pr-new@0.0 publish --pnpm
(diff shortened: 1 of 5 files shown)


AdrianGonz97:
just realized all packages that are a dependent of the modified package should also be published - working on it

AdrianGonz97:
yea, nevermind. the more i think about it, the more complicated it gets. let's just publish them all

AdrianGonz97:
everything looks to be working now 😅

benmccann:
wow, that's dedication!

manuel3108:
That's amazing!
I've just added a small comment to the top of the new file, so that in a year we know why we added this 😄

PR #412, opened 2024-06-17
chore: fix all type imports and exports

AdrianGonz97:
This PR enables [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) which will mainly be helpful for us in catching cases where we're not properly importing and exporting types

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,9 +8,11 @@
         "resolveJsonModule": true,
         "skipLibCheck": true,
         "declaration": true,
-        "allowSyntheticDefaultImports": true,
         "sourceMap": true,
         "strict": true,
-        "outDir": "build/"
+        "outDir": "build/",
+        "esModuleInterop": true,
+        "verbatimModuleSyntax": true,
+        "isolatedModules": true
     }
 }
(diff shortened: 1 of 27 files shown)


tsconfig.json
```diff
@@ -8,9 +8,11 @@
         "resolveJsonModule": true,
         "skipLibCheck": true,
         "declaration": true,
-        "allowSyntheticDefaultImports": true,
         "sourceMap": true,
         "strict": true,
-        "outDir": "build/"
+        "outDir": "build/",
+        "esModuleInterop": true,
```
benmccann, line 14: I think we shouldn't need this if we're outputting esm?

benmccann (commented):
cool! good changes!

we might also want to switch `moduleResolution` from `node` to `node16`

tsconfig.json
AdrianGonz97: we do need it for interoping with dependencies that use CJS, like `preferred-pm`

benmccann (approved):

AdrianGonz97:
> we might also want to switch `moduleResolution` from `node` to `node16`

that was part of my initial attempt, but there are quite a number of places where we're not abiding by the newer resolution algorithms, for example: [here](https://github.com/svelte-add/svelte-add/blob/f98454e93d84fc743a149c58428d841bec1d7bc2/packages/dev-utils/utils/update-packages.ts#L3-L5)

that can definitely be tackled in a separate PR though

manuel3108 (approved):

PR #414, opened 2024-06-18
chore: use esbuild to compile TS, update `exports` maps, and better externalized dependencies

AdrianGonz97:
This PR scratches several itches that I've had regarding TS compilation speeds, the mix-n-match of JSDoc & TS in the repo, wanting to adjust how externalized dependencies are determined, and making the `exports` maps for each package more uniformed.

## TS compilation speeds
According to [this comment](https://github.com/svelte-add/svelte-add/blob/21cfd6a0d9b149fcadce0f41e3f80e4a35d00cc7/rollup.config.js#L32-L39), we use a mix of JSDoc and TS in our adders and packages due to _super slow_ build times. 

I initially tried to outright replace `rollup` for `esbuild`, but I wasn't satisfied with the resulting code output, nor the inflexibility of its plugins. Instead, I found this awesome rollup plugin: [`rollup-plugin-esbuild`](https://github.com/egoist/rollup-plugin-esbuild), that uses `esbuild` only as a fast compiler to strip away all of the types, while maintaining the flexibility of `rollup`.

Now build times are even faster than it was before*!

\* initial build times when starting watch mode for rollup may be a tad slower (or so they may feel like it's slower because of generating `d.ts` files), but subsequent builds when files are changed are significantly faster

## Exports maps
I tweaked all of the `exports` maps for each package in their `package.json`s so that they're more uniformed. I also added a `files` field to each package so that the files in `build` are the only modules that are part of the distributed package (which should noticeably reduce the overall package size).

## Determining external dependencies
Rather than providing Rollup [a static array of dependencies](https://github.com/svelte-add/svelte-add/blob/21cfd6a0d9b149fcadce0f41e3f80e4a35d00cc7/rollup.config.js#L56) to externalize (to which we have to manually update every time we want to add a new dependency to it), externalized dependencies will now be determined based on the type of dependency they are, relative to their package:
- If the dependency is installed under `devDependencies`, then the dependency will be bundled
- If it's installed under `dependencies`, then it'll be externalized

I find this to be easier to reason about as well as to be less prone to errors, such as downloading the same thing twice. For example, consider the case of `preferred-pm`: https://github.com/svelte-add/svelte-add/blob/21cfd6a0d9b149fcadce0f41e3f80e4a35d00cc7/packages/core/package.json#L19-L24 
It's installed under `dependencies`, which means it will be downloaded as a transitive dependency when a user installs `@svelte-add/core`, but it's also bundled during builds because it's not externalized here: https://github.com/svelte-add/svelte-add/blob/21cfd6a0d9b149fcadce0f41e3f80e4a35d00cc7/rollup.config.js#L56
This means that it's _essentially_ being downloaded twice, which is not very efficient! This new method fixes that so that we don't even have to think about it anymore.

## Other small changes
- I took the liberty of removing some dependencies that we weren't using anywhere or were seemingly added by mistake (such as the case of `autoprefixer` being installed in the root).
- I also added the `node:` directive to all of the standard node library imports that I could find (e.g. `fs` -> `node:fs`)


AdrianGonz97:
Converting to draft as `@rollup/plugin-typescript` is overwriting and generating js modules, despite being set to only emit declaration files..

package.json
```diff
@@ -9,13 +9,13 @@
         "prettier:fix": "pnpm prettier . --write",
         "eslint:check": "pnpm eslint .",
         "eslint:fix": "pnpm eslint --fix .",
-        "build:dev": "rollup -cw",
+        "build:dev": "rollup -c --watch",
```
manuel3108, line 12:
```suggestion
        "build:dev": "rollup --config --watch",
```
If we are already getting more explicit, let's also make that for the other argument

packages/cli/package.json
```diff
@@ -9,19 +9,12 @@
...
-    "peerDependencies": {
-        "@svelte-add/bootstrap": "workspace:^",
-        "@svelte-add/bulma": "workspace:^",
-        "@svelte-add/mdsvex": "workspace:^",
-        "@svelte-add/routify": "workspace:^",
-        "@svelte-add/tailwindcss": "workspace:^",
-        "@svelte-add/threlte": "workspace:^",
-        "@svelte-add/storybook": "workspace:^",
-        "svelte-add": "workspace:^"
-    },
```
manuel3108, line 24: It's been a long time since I wrote this, but if I recall correctly this was added to make the changeset's work as expected. I initially added them as devDependencies so that they can properly be bundled, but [`changesets` refuse to work with those](https://github.com/changesets/changesets/issues/944)

At that time I couldn't find any proper workaround in the mentioned issue, and it does not look like someone has figured out something yet.

manuel3108 (approved):
I really like all the changes you have done! And the transpilation speed especially of the adders is crazy. 

The only thing I was wondering if it would be possible to generate the output and the .d.ts files in one go, with redefining the basically the same config in rollup multiple times. Was this not working because of incompatibilities between esbuild and typescript plugins? Am I overlooking something?

packages/cli/package.json
AdrianGonz97: makes sense!

AdrianGonz97:
> The only thing I was wondering if it would be possible to generate the output and the .d.ts files in one go, with redefining the basically the same config in rollup multiple times. Was this not working because of incompatibilities between esbuild and typescript plugins? Am I overlooking something?

I tried, though the `dts` plugin doesn't play well with the others like `nodeResolve`.

manuel3108:
I'm getting two new warnings now, are you expecting them?

Happens for normal and dts entry
```
bundles ./packages/core/index.ts → packages/core/build...
(!) Unresolved dependencies
https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
node:child_process (imported by "packages/core/utils/common.ts") 
```

and this happens only for dts

```
bundles ./packages/cli/index.ts → packages/cli/build...
(!) Generated an empty chunk
"index"
```

AdrianGonz97:
> I'm getting two new warnings now, are you expecting them?
> 
> Happens for normal and dts entry
> 
> ```
> bundles ./packages/core/index.ts → packages/core/build...
> (!) Unresolved dependencies
> https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
> node:child_process (imported by "packages/core/utils/common.ts") 
> ```

I'm not exactly sure why `rollup-plugin-dts` is unable to resolve it as a node module, but it should be fine since it's externalizing the module anyway.

> and this happens only for dts
> 
> ```
> bundles ./packages/cli/index.ts → packages/cli/build...
> (!) Generated an empty chunk
> "index"
> ```

This one is expected because the main CLI is both a lib (provides an `exports` field for `/website`) and an executable (uses the `bin` field). The empty chunk is just for the executable portion.

manuel3108:
> I'm not exactly sure why `rollup-plugin-dts` is unable to resolve `node:child_process` as a node module, but it should be fine since it's externalizing the module anyway.

But I'm getting that output twice. Or is that potentially because of the two entry points? That could explain it!

AdrianGonz97:
> Or is that potentially because of the two entry points? That could explain it!

Yep! it's because of the two entry points have their own configs and are handled separately

PR #415, opened 2024-06-18
feat: remove threlte adder in favor of simplified threlte setup instructions

benmccann:
The `threlte` setup instructions currently have multiple steps, but really shouldn't and it should just be as simple as `npm install threlte`. They've added extra steps as workarounds for bugs, but I don't think we should encourage that being the use case for `svelte-add`

- I've sent a PR to remove the unnecessary set of setting `"moduleResolution": "bundler"` in your `tsconfig.json`: https://github.com/threlte/threlte/pull/988
- I've also sent a fix to avoid the need for setting `ssr.noExternal` in your `vite.config.js`: https://github.com/threlte/threlte/pull/987

Ultimately I don't think `svelte-add` should have a built-in adder for every single Svelte library out there. We'd get hundreds of PRs (over 400 packages have sent PRs to be listed on https://www.sveltesociety.dev/packages). It's really meant for tooling and advanced application integrations like CSS processors, databases, authentication libraries, etc. It also just seems like something that 99% of apps aren't going to use.

I love threlte, but most apps don't use it or have 3d animations. This would probably be better as a community adder that they maintain if they'd like to keep it. They've already got their own `create-threlte` CLI which is far more powerful than the adder here, so folks would be much better served by using the solution they've provided


benmccann:
The new threlte docs are live with all the extra steps removed! https://threlte.xyz/docs/learn/getting-started/installation

PR #416, opened 2024-06-19
fix: improve formatting of `.env` file

benmccann:
The `.env` file is being spit out as:
```

DATABASE_URL="postgres://root:mysecretpassword@localhost:5432/local"
```

Better would be:
```
DATABASE_URL="postgres://root:mysecretpassword@localhost:5432/local"

```

Having a new line at the end is helpful in case people `cat` the file and it's also encouraged on GitHub where the user might check in an `.env.example` file


manuel3108 (changes requested):
This does not work as expected. The file is now generated as 

```

DATABASE_URL=file:local.db

```

benmccann:
Thanks for testing @manuel3108! I just fixed the merge conflict and then tested this PR manually locally. I wasn't able to reproduce it - though I did find it was inserting an extra newline and I fixed that. Could you let me know if it's still happening for you? If so, can you share the steps to reproduce?

manuel3108 (approved):
Works as expected now!

PR #421, opened 2024-06-20
feat: Single screen CLI

### LINKED ISSUE #410: Single screen CLI
### Topic

svelte-add

### Description

I'd had an idea for including headers in the CLI select described in https://github.com/bombshell-dev/clack/issues/197. That can't be implemented without support in `@clack/prompts`, so I've been trying to think how we can get closest to that with the current CLI APIs.

This would help group items without adding screens and as well as help users see very quickly what type of thing each library does.

An idea I had that I think could be cool would be to make the CLI look something like:
```
┌  Welcome to Svelte Add
│
◇  What would you like to add to your project?
│  ◻ CSS              | Bootstrap
│  ◻ CSS              | Bulma
│  ◻ CSS              | TailwindCSS
│  ◻ DB               | Drizzle
│  ◻ Markdown         | mdsvex
│  ◻ UI development   | Storybook
└
```

AdrianGonz97:
closes #410

This PR also adds `clack-core` and `clack-prompts` (which implements inactive group headers) to the repo as a temporary solution until https://github.com/bombshell-dev/clack/pull/199 is (hopefully!) merged.

Here's what the main CLI looks like now:
img
Now that the selection is flat, we should also probably categorize them a bit differently and more specific as well, such as: CSS, DB, Markdown, etc.

### DIFF
--- a/packages/clack-prompts/package.json
+++ b/packages/clack-prompts/package.json
@@ -1,0 +1,31 @@
+{
+    "name": "@svelte-add/clack-prompts",
+    "version": "0.9.0",
+    "type": "module",
+    "module": "./build/index.js",
+    "exports": {
+        ".": {
+            "types": "./build/index.d.ts",
+            "default": "./build/index.js"
+        },
+        "./package.json": "./package.json"
+    },
+    "types": "./build/index.d.ts",
+    "bugs": "https://github.com/svelte-add/svelte-add/issues",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/svelte-add/svelte-add/tree/main/packages/clack-prompts"
+    },
+    "files": [
+        "build"
+    ],
+    "license": "MIT",
+    "dependencies": {
+        "@svelte-add/clack-core": "workspace:*",
+        "sisteransi": "^1.0.5"
+    },
+    "devDependencies": {
+        "picocolors": "^1.0.0",
+        "is-unicode-supported": "^1.3.0"
+    }
+}
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -50,6 +50,8 @@ export default tseslint.config(
             "packages/website/.svelte-kit",
             "packages/website/build",
             "packages/website",
+            "packages/clack-core",
+            "packages/clack-prompts",
             "temp",
         ],
     },
--- a/packages/clack-core/CHANGELOG.md
+++ b/packages/clack-core/CHANGELOG.md
@@ -1,0 +1,199 @@
+# @clack/core
+
+## 0.4.0
+
+### Minor Changes
+
+- 127b7f0: feat: added `selectableGroups` option for inactive group headers for `GroupMultiSelectPrompt`
+
+## 0.3.4
+
+### Patch Changes
+
+- a04e418: fix(@clack/core): keyboard input not working after await in spinner
+- 4f6fcf5: feat(@clack/core): allow tab completion for placeholders
+
+## 0.3.3
+
+### Patch Changes
+
+- cd79076: fix: restore raw mode on unblock
+
+## 0.3.2
+
+### Patch Changes
+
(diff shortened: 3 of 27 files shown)


benmccann:
Yeah, making the categories something like CSS, DB, Markdown, UI development was the first thing that jumped out at me. I don't know if having a bunch of single-itemed categories will feel weird? The other way we might be able to handle that would be to collapse it if there's a single item like:
```
│  ◻ Database: Drizzle
```
Instead of:
```
│ Database
│  ◻ Drizzle
```

It's hard to know though without seeing it

This is awesome!!

packages/clack-prompts/package.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "@svelte-add/clack-prompts",
+    "version": "0.9.0",
```
benmccann, line 3: it might be nice to stick `"private": "true"` on these two new packages to prevent them from being published and have `@svelte-add/core` bundle them that way we don't have to pollute npm with what are hopefully short-lived temporary packages

AdrianGonz97:
> I don't know if having a bunch of single-itemed categories will feel weird? The other way we might be able to handle that would be to collapse it if there's a single item like:
> 
> ```
> │  ◻ Database: Drizzle
> ```
> 
> Instead of:
> 
> ```
> │ Database
> │  ◻ Drizzle
> ```

I think mixing and matching groups in that way might be more confusing 😅 , especially when it would line up along with whichever group is above it:
```
┌  Welcome to Svelte Add!
│
◆  What would you like to add to your project?
│   CSS
│  │ ◻ Bootstrap (Build fast, responsive sites with Bootstrap)
│  │ ◻ Bulma
│  │ ◻ TailwindCSS
│  └ ◻ Database: Drizzle
│   Markdown
│  └ ◻ mdsvex
│   Tools
│  └ ◻ Storybook
└
```

to make it _not_ part of any group would require more customizing of `@svelte-add/clack-prompts` (and _maybe_ `clack-core`), which might turn it into a permanent package for us since i don't really see a custom feature like that being added to clack

> It's hard to know though without seeing it

here's a preview with single item groups
img
I don't think it's _too_ bad

benmccann:
I think it'd probably look more like:
```
┌  Welcome to Svelte Add!
│
◆  What would you like to add to your project?
│   CSS
│  │ ◻ Bootstrap (Build fast, responsive sites with Bootstrap)
│  │ ◻ Bulma
│  └ ◻ TailwindCSS
│   Database: Drizzle
│   Markdown: mdsvex
│   UI development: Storybook
└
```

But yeah, actually seeing that written out in full I don't like it as much 

And I think the version with the single groups in your preview looks quite good actually! Plus, I imagine we'll end up with more multi-item groups over time

packages/clack-prompts/package.json
manuel3108: That sounds smart.

We should also add them here, otherwise the new packages will still be published I think. I'm not sure about this one, but at least that would explain why I added all those packages there initially:
https://github.com/svelte-add/svelte-add/blob/e7f414702ba319a649a22d4f73c4efc48d924ec6/package.json#L7

manuel3108 (changes requested):
I really like the changes. 

Additionally I don't mind the single grouped options as well. That's a massive improvement to the current situation

packages/clack-prompts/package.json
AdrianGonz97: heh, i actually haven't noticed the existence of that script until just now 😆

i don't think the publish script needs to be that complex. based on other projects i've setup changesets in, it should already respect the `"private": "true"` field, preventing the package from publishing

a simple `"ci:publish": "changesets publish"` script instead should do the job

manuel3108 (approved):
Sometimes things are just to obvious to see them yourself. Using `changeset publish` seem pretty smart. Let's give it a try!

PR #422, opened 2024-06-20
chore(pkg-pr-new): include peer deps in dependents check

### LINKED ISSUE #419: `pkg-pr-new` not releasing all required packages
### Topic

ci

### Version

latest

### Expected behavior

See #416 

We only published the drizzle adder. But when you try to execute the adder with someting like this: 

```
pnpm dlx https://pkg.pr.new/svelte-add/svelte-add@416 --path ./temp
```

you will get this error:
```
 ERR_PNPM_FETCH_404  GET https://pkg.pr.new/svelte-add/svelte-add@416: Not Found - 404
```

### Actual behavior

See #403 

we should also publish all (internal) dependencies of the project, in order to make that specific example work

AdrianGonz97:
closes #419

### DIFF
--- a/adders/drizzle/index.ts
+++ b/adders/drizzle/index.ts
@@ -6,4 +6,3 @@ import { checks } from "./config/checks.js";
 import { tests } from "./config/tests.js";
 
 export default defineAdder(adder, checks, tests);
-//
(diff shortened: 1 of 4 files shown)


manuel3108 (approved):
This seems to be working, even though the comment is not indicating that it also published `@svelte-add/core`. But that would be unnecessary in either case, so I will accept it like this 😄

PR #427, opened 2024-06-21
chore: remove readme symlink

manuel3108:
In v1 we had a symlink for the readme in place that somehow persisted until today. 
Earlier today I was working on my laptop and was running in codespaces, and they refused to open the readme, assumably because of the present symlink.

image

This (hopefully) removes the symlink, by deleting the file, committing, and re-adding the file. The [branch view inside github](https://github.com/svelte-add/svelte-add/tree/remove-readme-sysmlink) is at least not showing the symlink anymore.


manuel3108:
Second commit should of course be called "re-add readme", but we are squashing so this does not matter anyway.

benmccann (approved):
seems like a good idea!

PR #433, opened 2024-06-21
chore: Update `svelte-add` README.md

manuel3108:


packages/cli/README.md
```diff
@@ -22,4 +22,4 @@ If you want to install multiple adders at once, you can use this command:
 npx svelte-add@latest bulma mdsvex
```
benmccann, line 22: maybe we should demo tailwind instead of bulma? :laughing:

PR #434, opened 2024-06-21
chore: update adder descriptions

benmccann:
Updating the descriptions with details the differentiate between the options and help you choose. Like "The modern CSS framework that just works" is something that I imagine any CSS framework would say about themselves and I've honestly got no idea what "Headless" means in the context of an ORM.

### DIFF
--- a/.changeset/tough-dolls-wave.md
+++ b/.changeset/tough-dolls-wave.md
@@ -1,0 +1,10 @@
+---
+"@svelte-add/tailwindcss": patch
+"@svelte-add/bootstrap": patch
+"@svelte-add/storybook": patch
+"@svelte-add/drizzle": patch
+"@svelte-add/mdsvex": patch
+"@svelte-add/bulma": patch
+---
+
+chore: update adder description
--- a/adders/bootstrap/config/adder.ts
+++ b/adders/bootstrap/config/adder.ts
@@ -7,7 +7,7 @@ export const adder = defineAdderConfig({
     metadata: {
         ...generateAdderInfo(pkg),
         name: "Bootstrap",
-        description: "Build fast, responsive sites with Bootstrap",
+        description: "Sass-based CSS framework. 31.0% user retention - stateofcss.com",
         category: categories.css,
         environments: { kit: true, svelte: true },
         website: {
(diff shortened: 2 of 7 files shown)


manuel3108:
> is something that I imagine any CSS framework would say about themselves

Yeah, that's because that's where those texts came from, from their respective websites.

Not sure about the retention thing. Is this really something the user will use to differentiate the tools? But I understand that at least `bootstrap` and `bulma` would be the same without that.

benmccann:
I do look at those surveys when choosing tools and is one of the things that helped encourage me to use Svelte. But I recognize there are a lot of inputs to choose tools and this is only one source. Though their ranking does correlate quite highly with what I've heard from other maintainers/ambassadors. I think it would be helpful to have some way to indicate to novice users that we wouldn't necessarily recommend Bootstrap for a new project. But right now it's in the first position, which makes it seem favored.

Screenshot from 2024-06-21 13-20-32

AdrianGonz97:
I do agree with @manuel3108 that including the retention stats does feel a bit off

we could just express our recommendations by altering the list's order and then, instead of using descriptions, maybe we could link back to their respective docs pages (or maybe even a future svelte/svelte-add site pages that gives more detailed descriptions)? 

something like this:
img

alternatively, since we now control `@svelte-add/clack-prompts`, we could slightly alter the colors to reflect the most used packages as well:
img

or:
img
(it doesn't have to be green, but just for demoing purposes)

manuel3108:
(ordering of those adders is still on my list)

benmccann:
I removed the retention info, but I actually like @AdrianGonz97's idea of just putting a link to the site or docs, so perhaps I should remove the descriptions? Or do you want to keep them for now for the website?

AdrianGonz97:
> perhaps I should remove the descriptions? Or do you want to keep them for now for the website?

i think we can keep the descriptions and just change what's displayed in the prompt inside of `@svelte-add/core`  here:
https://github.com/svelte-add/svelte-add/blob/18080e5fb9c7240da70b100b46936ce0e81733c8/packages/core/adder/execute.ts#L221

benmccann:
thanks for the feedback! closing in favor of https://github.com/svelte-add/svelte-add/pull/435

Issue #432, opened 2024-06-21
`svelte-add` displays empty categories in cli


manuel3108:
### Topic

svelte-add

### Version

2.2.0

### Expected behavior

If a category contains 0 elements, it should not be displayed.

### Actual behavior

Currently it is displayed, especially if you create a svelte project and in example tried to run the `drizzle` adder
![image](https://github.com/svelte-add/svelte-add/assets/30698007/fa5048dc-4637-4a20-8c64-4c647a9aaa95)

PR #437, opened 2024-06-22
feat: custom ordered adders

manuel3108:
This prepares `svelte-add` to provide a non-alphabetical order of adders in the cli and the website. Additionally, this prepares a few types and files to support community adders, but no functionality at this point.

Notable changes
* There is now a package `@svelte-add/config` which provides the following
  * A list of categories extracted from `@svelte-add/core` (core was never really using the categories, only the cli and the website)
  * A mapping from categories to adders present in that category. The order the adders are written in this config, will be the order they will be displayed in the cli and on the website.
  * A list of community adders (not in use at this point, but I already added them since I was already generating this new package)
* Changed responsibilities for handling of categories
  * `@svelte-add/config` is now responsible for this
  * Now that we define categories and the adder display orders in a new package, we need to make sure to bump only required dependencies when adding new categories or changing display orders.
  * That's why `@svelte-add/config` is only used by `svelte-add` and `@svelte-add/website`. This avoids unnecessary bumping of `@svelte-add/core`
* Changed responsibilities for selecting adders to be applied. This was previously done by `@svelte-add/core` and is now done by `svelte-add` as this is the only place this was used. See version bumping reasoning before
* Rollup is not dynamically injecting a list of adders into `svelte-add` anymore. The list of supported adders is now determined by the adder category mapping.

Other changes
* Updated the ordering of the adders in the CSS category as discussed

To be discussed
* Naming of `@svelte-add/config` could be changed. 

Blocks #432

### DIFF
--- a/packages/config/adders/community.ts
+++ b/packages/config/adders/community.ts
@@ -7,11 +7,4 @@ export type CommunityAdder = {
 
 export type CommunityAdders = CommunityAdder[];
 
-export const communityAdders: CommunityAdders = [
-    {
-        id: "test",
-        displayName: "Test",
-        description: "Test description",
-        package: "@test/setup",
-    },
-];
+export const communityAdders: CommunityAdders = [];
--- a/packages/config/adders/community.ts
+++ b/packages/config/adders/community.ts
@@ -1,0 +1,17 @@
+export type CommunityAdder = {
+    id: string;
+    displayName: string;
+    description: string;
+    package: string;
+};
+
+export type CommunityAdders = CommunityAdder[];
+
+export const communityAdders: CommunityAdders = [
+    {
+        id: "test",
+        displayName: "Test",
+        description: "Test description",
+        package: "@test/setup",
+    },
+];
--- a/packages/config/adders/official.ts
+++ b/packages/config/adders/official.ts
@@ -1,0 +1,10 @@
+import type { AdderCategories } from "../categories";
+
+export const adderCategories: AdderCategories = {
+    css: ["tailwindcss", "bulma", "bootstrap"],
+    db: ["drizzle"],
+    markdown: ["mdsvex"],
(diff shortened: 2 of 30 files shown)


packages/config/adders/community.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type CommunityAdders = CommunityAdder[];
+
+export const communityAdders: CommunityAdders = [
+    {
+        id: "test",
+        displayName: "Test",
+        description: "Test description",
+        package: "@test/setup",
+    },
```
AdrianGonz97, line 16: should this be here or is this a leftover during testing?

AdrianGonz97 (dismissed):
I pushed a few minor tweaks, but overall I like the changes!

> To be discussed
> 
> * Naming of `@svelte-add/config` could be changed.

I think `config` is fine as a catchall term for where the adders are handled

packages/config/adders/official.ts
```diff
@@ -0,0 +1,10 @@
+import type { AdderCategories } from "../categories";
+
+export const adderCategories: AdderCategories = {
+    css: ["tailwindcss", "bulma", "bootstrap"],
+    db: ["drizzle"],
+    markdown: ["mdsvex"],
+    tools: ["storybook", "routify"],
```
benmccann, line 7: I wonder about making this something like `router` and `uiDevelopment` as tools is a bit generic and could become a catch-all

packages/config/adders/community.ts
manuel3108: Good catch, removed

packages/config/adders/official.ts
manuel3108: I couldn't imagine any other adders getting into those categories at this point. We can always change that later!

PR #440, opened 2024-06-23
fix: `docker-compose.yml` intendation

### LINKED ISSUE #418: Use `dedent`
### Topic

drizzle

### Version

`main`

### Expected behavior

no extra indentation or newlines

provide a `dedent` util similar to the one that kit uses in `core`. then use it in the drizzle adder and any others that need it

https://github.com/search?q=repo%3Asveltejs%2Fkit%20dedent&type=code

### Actual behavior

`docker-compose.yml` is highly indented:
```

                services:
                  db:
                    image: postgres
                    restart: always
                    ports:
                      - 5432:5432
                    environment:
                
                      POSTGRES_USER: root
                      POSTGRES_PASSWORD: mysecretpassword
                      POSTGRES_DB: local

```

it looks like others are affected as well

manuel3108:
Closes #418 

The original issue is only reproducible for me, when creating a new project via our cli, outside the monorepo. The `docker-compose.yml` was intended way to much in that case, and contained a new line that shouldn't be there.

This does not happen on pre-existing projects, as we try to run the locally installed prettier instance and try to format that file.

### DIFF
--- a/.changeset/cyan-dogs-give.md
+++ b/.changeset/cyan-dogs-give.md
@@ -2,4 +2,4 @@
 "@svelte-add/drizzle": patch
 ---
 
-fix: `docker-compose.yml` intendation
+fix: `docker-compose.yml` indentation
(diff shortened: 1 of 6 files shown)


AdrianGonz97 (approved):
what a nifty little utility! everything else looks good

PR #443, opened 2024-06-24
chore(CLI): adjust group header alignment

### LINKED ISSUE #442: Align lines and letters in CLI
### Topic

svelte-add

### Version

2.3.0

### Expected behavior

I did this in photoshop, so don't have the code for it unfortunately :laughing: 

aligned

### Actual behavior

Screenshot from 2024-06-24 12-19-33

AdrianGonz97:
closes #442

This alignment is definitely more pleasing to look at!

demo:
img

### DIFF
--- a/.changeset/ninety-kangaroos-hide.md
+++ b/.changeset/ninety-kangaroos-hide.md
@@ -2,4 +2,4 @@
 "svelte-add": patch
 ---
 
-chore: left align group headers
+chore: adjust group headers alignment
--- a/.changeset/ninety-kangaroos-hide.md
+++ b/.changeset/ninety-kangaroos-hide.md
@@ -1,0 +1,5 @@
+---
+"svelte-add": patch
+---
+
+chore: left align group headers
--- a/packages/clack-core/src/prompts/prompt.ts
+++ b/packages/clack-core/src/prompts/prompt.ts
@@ -227,7 +227,7 @@ export default class Prompt {
         if (diff) {
             const diffLine = diff[0];
             const lines = frame.split("\n");
-            let newLines = [];
+            let newLines: string[] = [];
 
             // If we don't have enough vertical space to print all of the lines simultaneously,
             // then we'll sticky the prompt message (first 3 lines) to the top so it's always shown.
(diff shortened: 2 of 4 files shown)


benmccann (approved):

AdrianGonz97:
here's the new look with the bars removed (single spaced):

img

---

and here's doubled spaced (the indentation is the same as if the bar was present):

img

though it does look a bit funny

benmccann:
this is great! I agree I prefer the single space version as well

AdrianGonz97:
updated to use the newline spacing:

img

i'm still not particularly a fan of it, but if the majority prefer it, then I really don't mind adding it

manuel3108:
> this is great! I agree I prefer the single space version as well

Agree!

> i'm still not particularly a fan of it, but if the majority prefer it, then I really don't mind adding it

Yeah, same for me. I understand both sides. As i'm personally always using the vscode terminal with minimal height, the prompts quickly start to be a bit buggy.  

https://github.com/svelte-add/svelte-add/assets/30698007/00389b38-e06e-4236-9004-d7cda7b6d230

But maybe I'm the only one having such a setup. I was thinking of providing a search as in the `changeset` cli, but that's a completely different topic.

https://github.com/svelte-add/svelte-add/assets/30698007/6a6c2038-a252-47c2-99da-d743a3ecf805

AdrianGonz97:
> Yeah, same for me. I understand both sides. As i'm personally always using the vscode terminal with minimal height, the prompts quickly start to be a bit buggy.

great point! we should fix this bug in `clack-core` before merging this PR since the additional vertical spacing makes it way more likely that users will experience this annoying issue

> But maybe I'm the only one having such a setup.

you're certainly not alone! 90% of the time, I'm working off of the embedded vscode terminal 😄

manuel3108:
> great point! we should fix this bug in `clack-core` before merging this PR since the additional vertical spacing makes it way more likely that users will experience this annoying issue

Just had a look through the `clack` repo, but it does not look like anyone has reported this issue, or opened a PR.

AdrianGonz97:
> Just had a look through the `clack` repo, but it does not look like anyone has reported this issue, or opened a PR.

they have, it just doesn't have the most descriptive name: https://github.com/bombshell-dev/clack/issues/116

and yea, no PRs have been raised for it either sadly

AdrianGonz97:
@manuel3108 I think I finally fixed it! this PR should be good to go now

manuel3108 (approved):
Perfect, works as expected now!

benmccann:
Awesome!!! Thanks for all your work on this!!

Would you be able to update your PR at `bombshell-dev/clack` with these changes too?

AdrianGonz97:
> Would you be able to update your PR at `bombshell-dev/clack` with these changes too?

I _may_ be able to with the version of this PR that only removes the `|` from the side, but I don't think the double-spacing or the change in text rendering that I did in `core` will be accepted over there 😅 (if it did, it should probably be a separate PR anyway).

Is the intention to use these changes in `create-svelte` as well? If so, we may have better luck (and more control) just publishing our version of clack prompts and core to npm

benmccann:
Hmm, maybe double spacing could be an option? Or apply only to the multiselect?

I shared an update on the progress this morning with some other maintainers and they did suggest we should try to avoid having two versions of the prompts library. I think the ideal would be to get our changes in upstream so we're not stuck maintaining our own version in the long-term. But if that can't happen or is taking too long then yeah we might have to publish our own version to npm and switch create-svelte to it

AdrianGonz97:
> Hmm, maybe double spacing could be an option? Or apply only to the multiselect?

sure, I can do that

> and they did suggest we should try to avoid having two versions of the prompts library. I think the ideal would be to get our changes in upstream so we're not stuck maintaining our own version in the long-term.

I agree, that would be perfect! 

however, to be frank, clack doesn't seem to be that well maintained. `@clack/prompts` hasn't received an update in [nearly a year](https://www.npmjs.com/package/@clack/prompts?activeTab=readme) while there's a plethora of [open PRs](https://github.com/bombshell-dev/clack/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen) for it (most of which are from a single contributor who keeps trying their best and not getting any responses 😭). I wouldn't hold my breath in hopes of any of them being merged anytime soon :/ 

checking out the maintainer's packages, they (relatively) recently reserved the name [`clink`](https://www.npmjs.com/package/clink), which looks like they're aiming to either replace `clack` or its current version (v1 rewrite maybe)? idk, it's hard to tell

all that being said, I don't think we should really be waiting on them, otherwise we may be waiting forever 😅

benmccann:
Interesting find! Thanks for updating the PR there. I agree we shouldn't block on them. If they haven't merged it by the time we get the adders ported over from `create-svelte` then we might as well publish our own version to npm

Issue #411, opened 2024-06-17
Decide how to have DB and auth adders interact with each other


benmccann:
### Topic

drizzle/lucia

### Description

I feel like the next adder we should add should potentially be for auth. Maybe Lucia (https://github.com/svelte-add/svelte-add/issues/297)

The auth providers and DB providers might interact with each other. E.g. to save auth info to the database. And Lucia has a Drizzle integration. But this provides a tricky challenge of what order to we run things in. Or what if a user runs one. And then later comes back and runs the other

### Describe the solution you'd like

We might have to define some sort of interface for these things to adhere to

manuel3108:
That's a pretty tough challenge as adders are planned to be run completely independently. 
We might be able to add some sort of config `runsBefore: ["lucia"],`. That way it would still be the job of core to handle those conflicts. 

If `drizzle` would run before `lucia` we could make the lucia adder check if `drizzle` is part of the dependencies inside the `package.json`. 

That could work without too much effort in my opinion, but might not be as flexible for the future.

benmccann:
Hmm, I just realized these aren't the only ones that will interact. I was thinking about porting over the ones from `create-svelte`, which are vitest, playwright, eslint, and prettier. There are some that interact, which can be seen here: https://github.com/sveltejs/kit/tree/main/packages/create-svelte/shared.

I think `create-svelte`'s setup has some nice properties, but I wouldn't necessarily do everything the way that is done there.  One thing it has that is really nice is that you can have a folder where any files in it are copied over including merging in the `package.json` in that directory.

One thing that might be easier with `svelte-add`'s code-based approach is some of the interactions like this one: https://github.com/sveltejs/kit/blob/main/packages/create-svelte/shared/%2Bplaywright%2Bvitest/package.json. That's not really a true interaction. It's just saying that if there's already a `test` field then run those existing tests in addition to the new ones. It's not necessarily a true interaction - though you would need that logic in both the playwright and vitest adders.

I think the one real interaction we have is that if you already have eslint or prettier and then run the other one then you should add `eslint-config-prettier`: https://github.com/sveltejs/kit/blob/5645614f497931f587b7cb8b3c885fce892a6a72/packages/create-svelte/shared/%2Beslint%2Bprettier/package.json#L3

[...]

PR #444, opened 2024-06-24
chore: remove `prettier` as a dependency

AdrianGonz97:
`prettier` is [ginormous](https://npmgraph.js.org/?q=svelte-add#select=exact%3Aprettier%403.3.2&sizing=) and is the largest dependency that users have to download every time they call any of our adders. 

Rather than having it as a direct dependency and using it's api in code, we can check if `prettier` is installed in their project and run the formatter via its CLI after all of the adders have executed and after dependency installations have occurred (covers the case of if `node_modules` isn't installed yet).

### DIFF
--- a/.changeset/selfish-dots-yell.md
+++ b/.changeset/selfish-dots-yell.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/core": patch
+---
+
+chore: removed `prettier` as a direct dependency and use the project's local installation of `prettier` for formatting
--- a/.changeset/wise-coins-hammer.md
+++ b/.changeset/wise-coins-hammer.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/ast-tooling": patch
+---
+
+fix: empty style and script tags are no longer included
--- a/adders/drizzle/config/tests.ts
+++ b/adders/drizzle/config/tests.ts
@@ -65,7 +65,7 @@ export const tests = defineAdderTests({
             contentType: "text",
             condition: ({ kit }) => kit.installed,
             content: ({ content }) => {
-                return content.replace("strict: true,", "");
+                return content.replace("strict: true", "");
             },
         },
         {
--- a/packages/ast-tooling/index.ts
+++ b/packages/ast-tooling/index.ts
@@ -140,19 +140,25 @@ export function serializeSvelteFile(asts: SvelteAst) {
     const css = serializePostcss(cssAst);
     const newScriptValue = serializeScript(jsAst);
 
-    const scriptTag = new Element("script", {}, undefined, ElementType.ElementType.Script);
-    for (const child of scriptTag.children) {
-        removeElement(child);
-    }
-    appendChild(scriptTag, new Text(newScriptValue));
-    appendChild(htmlAst, scriptTag);
+    if (newScriptValue.length > 0) {
+        const scriptTag = new Element("script", {}, undefined, ElementType.ElementType.Script);
+        for (const child of scriptTag.children) {
+            removeElement(child);
+        }
 
-    const styleTag = new Element("style", {}, undefined, ElementType.ElementType.Style);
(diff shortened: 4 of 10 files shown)


benmccann:
sounds like a good idea to me. I don't think it should affect this, but I'd like to move the `prettier` setup (along with all the others) from `create-svelte` to this repo. Rich confirmed today that he was okay with that

AdrianGonz97:
> sounds like a good idea to me. I don't think it should affect this, but I'd like to move the `prettier` setup (along with all the others) from `create-svelte` to this repo. Rich confirmed today that he was okay with that

cool! and yea, this PR wouldn't block the ability to add a prettier adder

manuel3108 (commented):
I like the idea, but I think there are a couple of problems at the current point in time.

Let's start by saying that we won't be able to format the files if the user decides to skip installing dependencies. That might cause some files to have a weird indentation or formatting. Although as of know, all of the adders to behave not to badly.

If you are creating a new project through `svelte-add` right now, the only way you will get prettier installed and thus your files formatted would be if you select to create a new kit project and additionally select to install add prettier and decided to install the dependencies.  The projects created by `create-vite` do not seem to bring prettier by default. In total I think that's a lot of conditions that need to apply to make this work. Looking into the future at the other hand, where we will provide the prettier adder as part of the `svelte-add` cli, that makes a ton more sense.

And for the last thing I noticed I'm not sure if it belongs here. If I create a new kit project via the cli inside our monorepo (I usually use `./temp`), select to add prettier, and install the dependencies, the following is printed to my console:
```
◇  Failed to format files
│
■  [error] Cannot find package 'prettier-plugin-tailwindcss' imported from D:\dev\web\svelte-add\temp\noop.js
│
│
└  You're all set!
```

If I execute exactly the same setup in a different folder outside of our repo root (let's say `../svelte-add-temp`), then this error is not shown and everything works as expected.

---
To summarize I think this change makes a ton of sense dependency wise. Though at this point merging this might look like a step back, because our own `prettier` adder is not ready yet. I'll leave it to you guys to decide if we should merge this at this point, as I really don't have a strong opinion in any direction right now.

AdrianGonz97:
> Let's start by saying that we won't be able to format the files if the user decides to skip installing dependencies. That might cause some files to have a weird indentation or formatting. Although as of know, all of the adders to behave not to badly.

I think having the adders format the modified files is more of a nice-to-have rather than a required feature. I don't think it's particularly worth it at the cost of doubling the total package size + dependencies for our adders.

Additionally, formatting for the users who _don't_ have `prettier` also has its own downsides as well. Most notably for those that are using different formatters (biomejs, eslint stylistic, etc). We're now prescribing a certain formatting style that likely doesn't conform to whichever a user has set and will probably have to be reformatted by the user anyway.

> And for the last thing I noticed I'm not sure if it belongs here. If I create a new kit project via the cli inside our monorepo (I usually use ./temp), select to add prettier, and install the dependencies, the following is printed to my console:

this looks like an edge-case because of monorepos, although I think I have an idea as to how to fix it

> Though at this point merging this might look like a step back, because our own prettier adder is not ready yet.

I do see your point with not having a prettier adder, though. I don't think it'll be that big of a lift to add it, so I can open a new PR to add it as a pre-req for this PR

manuel3108:
> Additionally, formatting for the users who _don't_ have `prettier` also has its own downsides as well. Most notably for those that are using different formatters (biomejs, eslint stylistic, etc). We're now prescribing a certain formatting style that likely doesn't conform to whichever a user has set and will probably have to be reformatted by the user anyway.

Havn't looked at it from this side, that's a strong point.

> I do see your point with not having a prettier adder, though. I don't think it'll be that big of a lift to add it, so I can open a new PR to add it as a pre-req for this PR

Let's take our time, as there are some issues that we need to solve before properly being able to manage this, see #445, #411

manuel3108 (approved):

Issue #445, opened 2024-06-24
core method to merge a template with project


benmccann:
### Topic

@svelte-add/core

### Description

`create-svelte` works by copying template files into the user's project. Things like an `.eslintignore`, `.prettierrc`, etc. get copied over and just dropping them in this directory can be a little nicer than inserting their contents into a method in a `.js` file. It would be nice to port over the prettier, eslint, playwright, and vitest setup into this project and having that functionality be available would make it a lot easier

### Describe the solution you'd like

Some features it has:
- you can put `+typescript` or `-typescript` in the directory name if it should run only if you do or don't have typescript installed. It also works with any adder name, but I'm not sure that's necessarily required. Just being able to have a TypeScript and non-TypeScript version would be a great start that handles most of the cases
- you can put a `package.json` in that directory and it will be merged into the user's `package.json`
- you can put sub-directories like `src/lib/db` or `src/routes/login` and it will copy files over with that layout. this is less used that the other two, but can be a nice thing to have

manuel3108:
I don't fully agree with this one yet. 

There is a few things that we need to keep in mind. `create-svelte` is intended to be a one-off command to kickof a project. Thus, you cannot execute `create-svelte` on an existing project, which is one of the reasons they come a away with having template files like in [`+eslint+prettier`](https://github.com/sveltejs/kit/tree/main/packages/create-svelte/shared/%2Beslint%2Bprettier), where they just assume that they can overwrite the `lint` command inside the `package.json` because it's been a new project, a nobody other than the cli should have been able to make modifications at this point.

Since our cli can be run even later on, it could already be the case that the user has already added a `lint` command, potentially with some completely unexpected tooling. That's why I don't think it will be safe enough to just merge files like the `package.json`. But I think that this would only apply for files that are modified quite frequently and do not belong to any tools we have adders for. That would mean that I think your idea could work perfectly for files like the `eslint.config.js`

For the mentioned `eslint.config.js` I could imagine having a file definition like this inside the adder definition
```js
{
    name: () => "eslint.config.js",
    contentType: "script",
    content: ({ typescript, fromTemplate }) => {
        fromTemplate(`eslint/${typescript.installed ? "+" : "-"}/eslint.config.js`);
    },
},
```

As for the `package.json` I think we could perfectly handle that with our current tooling without any additional work.

benmccann:
[...]

PR #446, opened 2024-06-25
chore: consolidate adders

### LINKED ISSUE #441: Consider not publishing each adder individually to npm
### Topic

all adders

### Description

Currently we are publishing all adders seperatly to npm (`@svelte-add/tailwindcss`). Additionally they are bundled into the `svelte-add` cli.

Initially this was done to provide an easy to use command to install a single adder. With some of the latest changes we have done lately, this has probably become way more unimportant. Let's compare the two possibilities to apply the tailwindcss adder.
```
pnpm dlx @svelte-add/tailwindcss@latest
```

vs.
```
pnpm dlx svelte-add@latest tailwindcss
```

As of now, both commands contain exactly the same words and have nearly exactly the same number of characters (second is one character shorter). To me the second one feels more natural and would also bring the benefit, that you already know the command for the interactive cli by heart, in case you need it at some point

```
pnpm dlx svelte-add@latest 
```

---

Additionally this could remove the need of the peer dependencies here and solve some of the challenges we had with changesets in the past: https://github.com/svelte-add/svelte-add/blob/b42e0a0169a717b9381843e0689d135a64c3f1f3/packages/cli/package.json#L20-L26

We will need to think of we can properly represent this in the repo, while still keeping some flexibility for #411 . We might want to create a new private package `@svelte-add/adders` that lives in the `./adders` directory. This will work because the adders are bundled into the cli in either way. 

The only thing I'm unsure about is if changesets would still work as we expect in this configuration. We might need to give it a try to find that out.

manuel3108:
Closes #441 

The following things need to be done:

- [ ] deprecate published adder npm packages (i.e. `@svelte-add/tailwindcss`
- [x] Some tests are failing (only one or two per adder)
- [x] Test if it's still possible to create independent community adders (requires modifications, but works)
- [x] Test if the script to update deps is still working as expected
- [x] Test if the cli is working as expected

Notable changes:
- We now have `@svelte-add/adders` instead of many different packages for all adders. This will also allow us to easily implement #411, #445 
- `@svelte-add/adders` is private and will not be published, only bundled by `svelte-add` (and the website)
- I reworked some of the internal dependencies, to remove dependencies from `svelte-add` and rather add them to `@svelte-add/config` where required. I also removed some useless internal dependencies.


packages/testing-library/utils/adder.ts
```diff
@@ -12,7 +12,11 @@ export async function runAdder(
     adderOptions[adder.config.metadata.id] = optionValues;
     const remoteControlOptions: RemoteControlOptions = { workingDirectory, isTesting: true, adderOptions };
 
-    await executeAdder({ config: adder.config, checks: adder.checks }, remoteControlOptions);
+    await executeAdder(
+        { config: adder.config, checks: adder.checks },
+        { name: "testing-cli", version: "tbd" },
```
AdrianGonz97, line 17: is this a leftover?

AdrianGonz97 (dismissed):
everything is looking, as well as working, great!

and this is always a lovely thing to see 😄:
img

packages/testing-library/utils/adder.ts
manuel3108: This generally needs to be there, but I have adjusted the naming to better reflect the package name

AdrianGonz97:
It looks like the newly released version of `pkg-pr-new@0.0.10` is throwing some errors. In `scripts/get-deps-to-publish.js`, can you pin the version by updating `execSync` to:
 ```
execSync(`pnpm dlx pkg-pr-new@0.0.9 publish --pnpm ${paths}`, { stdio: "inherit" });
```

manuel3108:
> It looks like the newly released version of `pkg-pr-new@0.0.10` is throwing some errors. In `scripts/get-deps-to-publish.js`, can you pin the version by updating `execSync` to:
> 
> ```
> execSync(`pnpm dlx pkg-pr-new@0.0.9 publish --pnpm ${paths}`, { stdio: "inherit" });
> ```

Yeah, I have just raised an issue upstream. Will ignore it for now, as we are done this PR and they will probably fix it in a little bit I assume

PR #449, opened 2024-06-29
chore: fix changesets

AdrianGonz97:
scripts `pnpm utils:readmes && pnpm utils:packages` no longer exists and is causing the [changesets workflow to fail](https://github.com/svelte-add/svelte-add/actions/runs/9725879570).

`ci:version` is also no longer needed since we don't have to do any additional processing before versioning the packages. the changesets action runs `changesets version` by default, so we can safely remove it.


manuel3108 (approved):
Good catch!

PR #451, opened 2024-06-30
feat: `prettier` and `eslint` adders

manuel3108:
This PR aims to add prettier and eslint adders. As discussed in #445 both adders depend on each other in some way, that's why I have decided that it's a good idea to add them in the same PR.

We might need to also implement #445 or #411 here to make everything work as expected, but that remains to be seen.

At this point this PR is only a POC of the `prettier` adder.

### DIFF
--- a/adders/prettier/config/adder.ts
+++ b/adders/prettier/config/adder.ts
@@ -1,0 +1,59 @@
+import { dedent, defineAdderConfig } from "@svelte-add/core";
+import { options } from "./options";
+
+export const adder = defineAdderConfig({
+    metadata: {
+        id: "prettier",
+        name: "Prettier",
+        description: "An opinionated code formatter",
+        environments: { svelte: true, kit: true },
+        website: {
+            logo: "./prettier.svg",
+            keywords: ["prettier", "code", "formatter", "formatting"],
+            documentation: "https://prettier.io/docs/en/",
+        },
+    },
+    options,
+    integrationType: "inline",
+    packages: [
+        { name: "prettier", version: "^3.1.1", dev: true },
+        { name: "prettier-plugin-svelte", version: "^3.1.2", dev: true },
+    ],
+    files: [
+        {
+            name: () => `.prettierrc`,
+            contentType: "text",
+            content: () => {
+                return dedent`
+                {
+                    "useTabs": true,
+                    "singleQuote": true,
+                    "trailingComma": "none",
+                    "printWidth": 100,
+                    "plugins": ["prettier-plugin-svelte"],
+                    "overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
+                }`;
+            },
+        },
+        {
+            name: () => `.prettierignore`,
+            contentType: "text",
+            content: () => {
+                return dedent`
+                # Package Managers
+                package-lock.json
+                pnpm-lock.yaml
+                yarn.lock`;
+            },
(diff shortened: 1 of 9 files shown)


adders/prettier/config/adder.ts
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    "useTabs": true,
+                    "singleQuote": true,
+                    "trailingComma": "none",
+                    "printWidth": 100,
+                    "plugins": ["prettier-plugin-svelte"],
+                    "overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
+                }`;
+            },
+        },
```
AdrianGonz97, line 37: This should be a bit more robust if they already have a prettier config:
```suggestion
        {
            name: () => `.prettierrc`,
            contentType: "json",
            content: ({ data }) => {
                data.useTabs ??= true;
                data.singleQuote ??= true;
                data.trailingComma ??= "none";
                data.printWidth ??= 100;
                data.plugins ??= [];
                data.overrides ??= [];

                const plugins: string[] = data.plugins;
                if (!plugins.includes("prettier-plugin-svelte")) {
                    data.plugins.unshift("prettier-plugin-svelte");
                }

                const overrides: { files: string | string[]; options?: { parser?: string } }[] = data.overrides;
                const override = overrides.find((o) => o?.options?.parser === "svelte");
                if (!override) {
                    overrides.push({ files: "*.svelte", options: { parser: "svelte" } });
                }
            },
```

adders/prettier/config/adder.ts
```diff
@@ -0,0 +1,59 @@
...
+            name: () => `.prettierignore`,
+            contentType: "text",
+            content: () => {
+                return dedent`
+                # Package Managers
+                package-lock.json
+                pnpm-lock.yaml
+                yarn.lock`;
+            },
+        },
```
AdrianGonz97, line 48: I've done something similar, but perhaps we should consider not overwriting their file if it already exists:
```suggestion
        {
            name: () => `.prettierignore`,
            contentType: "text",
            content: ({ content }) => {
                if (content) return content;
                return dedent`
                    # Package Managers
                    package-lock.json
                    pnpm-lock.yaml
                    yarn.lock
                `;
            },
        },
```

adders/prettier/config/adder.ts
```diff
@@ -0,0 +1,59 @@
...
+        {
+            name: () => `package.json`,
+            contentType: "json",
+            content: ({ data }) => {
+                if (!data.scripts) data.scripts = {};
+
+                data.scripts.lint = "pnpm prettier . --check";
```
AdrianGonz97, line 55: this should be a bit more idiomatic, though `format:check` would only be included if `eslint` isn't detected:
```suggestion
                data.scripts ??= {};
                const scripts: Record<string, string> = data.scripts;
                scripts["format"] ??= "prettier --write .";
                scripts["format:check"] ??= "prettier --check .";
```

AdrianGonz97 (commented):
heh, I actually have my own implementation of the prettier adder before [this reply](https://github.com/svelte-add/svelte-add/pull/444#issuecomment-2189666418) was commented, so I've left some of our differences as suggestions below:

manuel3108:
Looks like your implementation is way more robust at this time. 

Closing this inf scout of your implementation. I will focus on #411 in the meantime

PR #453, opened 2024-06-30
feat: prettier adder

AdrianGonz97:
Currently a WIP, but opening the PR for now. Some things, like the svg and workspace reloading, have been taken from #451


AdrianGonz97:
A few notes:
- I've added a decent starting point for `eslint.config.js` modifications. It resides in the `prettier` adder for now, but once an `eslint` adder is implemented, it can be extracted out into somewhere where both adders can use it
- I've also renamed the new category from `devTools` to `common` and hoisted it to the top so that `prettier` and (when implemented) `eslint` can be accessed more easily as they are likely going to be the most commonly used adders

adders/prettier/config/adder.ts
```diff
@@ -0,0 +1,128 @@
...
+        },
+        {
+            name: () => "package.json",
+            contentType: "json",
+            content: ({ data, dependencies }) => {
+                data.scripts ??= {};
+                const scripts: Record<string, string> = data.scripts;
+                const CHECK_CMD = "prettier --check .";
+                scripts["format"] ??= "prettier --write .";
+                scripts["format:check"] ??= CHECK_CMD;
```
manuel3108, line 73: Do we need that? The current `create-svelte` is not adding that command, and you are able to run that by running the `lint` script
[-eslint+prettier](https://github.com/sveltejs/kit/blob/main/packages/create-svelte/shared/-eslint%2Bprettier/package.json)
[+eslint-prettier](https://github.com/sveltejs/kit/blob/main/packages/create-svelte/shared/%2Beslint-prettier/package.json)
[+eslint+prettier](https://github.com/sveltejs/kit/blob/main/packages/create-svelte/shared/%2Beslint%2Bprettier/package.json)

adders/prettier/config/adder.ts
```diff
@@ -0,0 +1,128 @@
...
+                    // append to the end as a fallback
+                    array.elements.push(...nodesToInsert);
+                }
+            },
+        },
+    ],
+});
+
+function hasEslint(deps: Record<string, string>): boolean {
+    return !!deps["eslint"] && deps["eslint"].startsWith("9");
```
manuel3108, line 127: What would we do if a user has v8 installed?
Should we maybe make this a pre-requisite and hint him to upgrade his eslint first?

adders/prettier/config/tests.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const tests = defineAdderTests({
+    files: [],
+    options,
+    optionValues: [],
+    tests: [
+        {
+            name: "todo",
+            run: async () => {},
+        },
```
manuel3108, line 12: I don't think we will be able to test this at all. The only think that could work here is using the postconditions. 

IIRC tests for an adder should fail, if a precondition fails

packages/config/categories.ts
```diff
@@ -4,12 +4,17 @@ export type CategoryInfo = {
...
-export type CategoryKeys = "css" | "tools" | "db" | "markdown";
+export type CategoryKeys = "common" | "css" | "db" | "markdown" | "tools";
 export type CategoryDetails = Record<CategoryKeys, CategoryInfo>;
 
 export type AdderCategories = Record<CategoryKeys, string[]>;
 
 export const categories: CategoryDetails = {
+    common: {
+        id: "common",
+        name: "Common Tooling",
```
benmccann, line 15: "Tools" and "Common Tooling" seem too similar as category names. perhaps this category could be something like "Code hygiene"?

packages/core/utils/workspace.ts
```diff
@@ -71,6 +72,11 @@ export async function populateWorkspaceDetails(workspace: WorkspaceWithoutExplic
         workspace.typescript.installed = "tslib" in packageJson.devDependencies;
         workspace.prettier.installed = "prettier" in packageJson.devDependencies;
         workspace.kit.installed = "@sveltejs/kit" in packageJson.devDependencies;
+        workspace.dependencies = { ...packageJson.devDependencies, ...packageJson.dependencies };
```
manuel3108, line 75: I'm honestly not sure if we should mix deps and dev deps here. But tbh, I cannot imagine any reasoning here why this should not be the case. So feel free to resolve this.

adders/prettier/config/adder.ts
```diff
@@ -0,0 +1,128 @@
...
+                    pnpm-lock.yaml
+                    yarn.lock
+                `;
+            },
+        },
+        {
+            name: () => ".prettierrc",
+            contentType: "json",
+            content: ({ data }) => {
+                data.useTabs ??= true;
```
benmccann, line 46: I'm not sure we should set these properties if the user already has a `.prettierrc`. It essentially means they've setup prettier already and we're modifying their config. I think we could add `prettier-plugin-svelte` in that case, but I don't think I'd change any formatting options

Maybe this section should be more like:
```
if (!Object.keys(data).length) {
  data.useTabs = true;
  data.singleQuote = true;
  data.trailingComma = 'none';
  data.printWidth = 100;
}
```

manuel3108: Isn't that code exactly doing this? At least from my c# backgound i would assume that this only assigns useTabs if useTabs is null / undefined

adders/prettier/config/adder.ts
AdrianGonz97: > What would we do if a user has v8 installed?

it just won't modify the config if it's v8. we should only be targeting v9 since the flat config is the new default for it. having to account for both config formats is something we definitely don't want to support 😄

> Should we maybe make this a pre-requisite and hint him to upgrade his eslint first?

I don't think we need a pre-req for this. although, a post-execution log could just say that "eslint v9 or higher wasn't 
 detected and X step was skipped" or something along those lines.

packages/core/utils/workspace.ts
AdrianGonz97: Yea I couldn't think of any scenario where being a dep or devDep would make a difference, so I merged them together for easier usage

adders/prettier/config/tests.ts
AdrianGonz97: I was thinking something along the lines of reading the config files and confirming they were updated/created

adders/prettier/config/adder.ts
AdrianGonz97: fair point! will address it

adders/prettier/config/tests.ts
manuel3108: Yeah, that could work

adders/prettier/config/adder.ts
AdrianGonz97: > I'm not sure we should set these properties if the user already has a `.prettierrc`. It essentially means they've setup prettier already and we're modifying their config. I think we could add `prettier-plugin-svelte` in that case, but I don't think I'd change any formatting options

Ah this is an excellent point, nice catch!

> Isn't that code exactly doing this? At least from my c# backgound i would assume that this only assigns useTabs if useTabs is null / undefined

yes and no. yes that it only assigned `useTabs` if it's `undefined`, but if they already have a prettier config configured, we don't want to add new rules to it (with the exception of adding the `prettier-plugin-svelte`)

benmccann: > this only assigns useTabs if useTabs is null / undefined

that's correct

> Isn't that code exactly doing this?

not quite. E.g. if the user has setup a `.prettierrc` which only specifies `singleQuote = true` and they've left `useTabs` empty, then we would come along and set `useTabs = true` and I'm not sure that's what we want

packages/config/categories.ts
AdrianGonz97: how about "Linters"?

adders/prettier/config/tests.ts
AdrianGonz97: hrmm, I'm running into a bit of a wall right now where I can't really read the contents of the `.prettierrc` because the `run` hook in tests doesn't provide a `cwd`. Perhaps we can save this for later until #461 is implemented?

packages/config/categories.ts
benmccann: To me, prettier is a formatter, which is a different category than linter. "Code quality" could be another option

adders/prettier/config/adder.ts
benmccann: That sounds good to me. Though "eslint v9 or higher wasn't detected and X step was skipped" might not be quite the right wording as it would work just fine when there's no eslint installed. Instead I think you'd have to say "An older major version of eslint was detected and step X was skipped"

packages/config/categories.ts
manuel3108: I like code quality tbh

adders/prettier/config/adder.ts
AdrianGonz97: How about this for the warning?:
img

Leaving out the eslint version does feel a bit off, how would the user know which one to upgrade to?

benmccann: that looks great. I think the message is fine as you have it because I would assume that means you should use the latest version. But if you wanted to include the versions you could say something like "eslint 8 was detected, but eslint 9+ is required."

manuel3108 (approved):
LGTM, works as expected.

PR #455, opened 2024-07-01
chore: update vscode settings

manuel3108:
After merging #452, my vscode was still complaining about some type errors, although `pnpm types:check` ran flawlessly. I figured out that my vscode was using the bundled typescript version which is not up to date.

While i was in that config i got a warning that i can remove the experimental key for eslint.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,5 @@
     "typescript.tsserver.experimental.enableProjectDiagnostics": true,
     "cSpell.words": ["ecma", "popperjs", "postcondition", "Postcondition", "postconditions", "Svex", "threlte"],
     "eslint.useFlatConfig": true,
-    "typescript.tsdk": "node_modules\\typescript\\lib"
+    "typescript.tsdk": "./node_modules/typescript/lib"
 }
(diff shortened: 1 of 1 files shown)


benmccann (approved):

.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "typescript.tsserver.experimental.enableProjectDiagnostics": true,
     "cSpell.words": ["ecma", "popperjs", "postcondition", "Postcondition", "postconditions", "Svex", "threlte"],
-    "eslint.experimental.useFlatConfig": true
+    "eslint.useFlatConfig": true,
+    "typescript.tsdk": "node_modules\\typescript\\lib"
```
AdrianGonz97, line 5: this path looks to be a windows specific as the filepath separator is using `\\`. is there a way to have these be dynamic to the current user's platform?

AdrianGonz97 (changes requested):

.vscode/settings.json
AdrianGonz97: does setting it to `node_modules/typescript/lib` work for windows?

AdrianGonz97: was able to get it to work on my machine! `/` is compatible with windows, so setting it to this should work for everyone:
```suggestion
    "typescript.tsdk": "./node_modules/typescript/lib"
```

manuel3108: Good catch, yes this does work!

PR #462, opened 2024-07-01
fix: only add a slot element if it's missing

### LINKED ISSUE #459: tailwind adder is broken
### Topic

tailwind

### Version

6.3.2

### Expected behavior

It should correctly update the layout file

### Actual behavior

Screenshot from 2024-07-01 13-28-22

AdrianGonz97:
fixes #459

so the main broken part is that another `<slot>` element is added when one already exists in the layout. this fixes that, but the formatting is still sorta ugly if `prettier` isn't installed. 

to (somewhat) make it less horrific without prettier formatting, I've also updated `serializeSvelteFile` so that the script tag is added at the top, rather than at the bottom with `<style>` .

so rather than this:
<details>

```svelte


<div class="app">
	<Header></Header>

	<main>
		<slot></slot>
	</main>

	<footer>
		<p>
			visit <a href="https://kit.svelte.dev">kit.svelte.dev</a> to learn SvelteKit
		</p>
	</footer>
</div>


<script>import "../app.css";
import Header from "./Header.svelte";
import "./styles.css";</script><style>
	.app {
		display: flex;
		flex-direction: column;
		min-height: 100vh;
	}

	main {
		flex: 1;
		display: flex;
		flex-direction: column;
		padding: 1rem;
		width: 100%;
		max-width: 64rem;
		margin: 0 auto;
		box-sizing: border-box;
	}

	footer {
		display: flex;
		flex-direction: column;
		justify-content: center;
		align-items: center;
		padding: 12px;
	}

	footer a {
		font-weight: bold;
	}

	@media (min-width: 480px) {
		footer {
			padding: 12px 0;
		}
	}
</style>
```

</details>

---

it'll now produces this:
<details>

```svelte
<script>import "../app.css";
import Header from "./Header.svelte";
import "./styles.css";</script>

<div class="app">
	<Header></Header>

	<main>
		<slot></slot>
	</main>

	<footer>
		<p>
			visit <a href="https://kit.svelte.dev">kit.svelte.dev</a> to learn SvelteKit
		</p>
	</footer>
</div>


<style>
	.app {
		display: flex;
		flex-direction: column;
		min-height: 100vh;
	}

	main {
		flex: 1;
		display: flex;
		flex-direction: column;
		padding: 1rem;
		width: 100%;
		max-width: 64rem;
		margin: 0 auto;
		box-sizing: border-box;
	}

	footer {
		display: flex;
		flex-direction: column;
		justify-content: center;
		align-items: center;
		padding: 12px;
	}

	footer a {
		font-weight: bold;
	}

	@media (min-width: 480px) {
		footer {
			padding: 12px 0;
		}
	}
</style>
```

</details>

it's not great, but it's better than what it was before

### DIFF
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -93,7 +93,7 @@ export const adder = defineAdderConfig({
                 js.imports.addEmpty(js.ast, "../app.css");
                 const slot = html.element("slot");
                 // @ts-expect-error should be compatible
-                if (js.common.hasNode(html.ast.childNodes, slot)) {
+                if (!js.common.hasNode(html.ast.childNodes, slot)) {
                     html.ast.childNodes.push(slot);
                 }
             },
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -92,8 +92,7 @@ export const adder = defineAdderConfig({
             content: ({ js, html }) => {
                 js.imports.addEmpty(js.ast, "../app.css");
                 const slot = html.element("slot");
-                // @ts-expect-error should be compatible
-                if (!js.common.hasNode(html.ast.childNodes, slot)) {
+                if (html.ast.childNodes.length === 0) {
                     html.ast.childNodes.push(slot);
                 }
             },
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -91,8 +91,8 @@ export const adder = defineAdderConfig({
             contentType: "svelte",
             content: ({ js, html }) => {
                 js.imports.addEmpty(js.ast, "../app.css");
-                const slot = html.element("slot");
                 if (html.ast.childNodes.length === 0) {
+                    const slot = html.element("slot");
                     html.ast.childNodes.push(slot);
                 }
             },
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -92,7 +92,10 @@ export const adder = defineAdderConfig({
             content: ({ js, html }) => {
                 js.imports.addEmpty(js.ast, "../app.css");
                 const slot = html.element("slot");
-                html.ast.childNodes.push(slot);
+                // @ts-expect-error should be compatible
+                if (js.common.hasNode(html.ast.childNodes, slot)) {
+                    html.ast.childNodes.push(slot);
+                }
             },
             condition: ({ kit }) => kit.installed,
         },
--- a/.changeset/brave-dolphins-guess.md
+++ b/.changeset/brave-dolphins-guess.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/adders": patch
+---
+
+fix: only add `slot` if the layout file is newly created
--- a/packages/ast-tooling/index.ts
+++ b/packages/ast-tooling/index.ts
@@ -2,7 +2,7 @@ import { parse as tsParse } from "recast/parsers/typescript.js";
 import { parse as recastParse, print as recastPrint } from "recast";
 import { Document, Element, Text, type ChildNode } from "domhandler";
 import { ElementType, parseDocument } from "htmlparser2";
-import { appendChild, removeElement, textContent } from "domutils";
+import { appendChild, prependChild, removeElement, textContent } from "domutils";
 import serializeDom from "dom-serializer";
 import { Root as CssAst, Declaration, Rule, AtRule, Comment } from "postcss";
 import { parse as postcssParse } from "postcss";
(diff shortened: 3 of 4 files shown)


adders/tailwindcss/config/adder.ts
```diff
@@ -92,7 +92,10 @@ export const adder = defineAdderConfig({
             content: ({ js, html }) => {
                 js.imports.addEmpty(js.ast, "../app.css");
                 const slot = html.element("slot");
-                html.ast.childNodes.push(slot);
+                // @ts-expect-error should be compatible
+                if (!js.common.hasNode(html.ast.childNodes, slot)) {
```
benmccann, line 96: why should we have to check if there's a slot at all?  I don't quite understand why we need to do anything more than add the stylesheet for a css integration

AdrianGonz97: > why should we have to check if there's a slot at all?

it covers the case of if the `+layout.svelte` isn't already present in the project, then the adder will create the layout file and add the `slot` element

having a second look now though, I was definitely overthinking this. If the `+layout.svelte` file is newly created, then `html.ast.childNodes` should be empty, so we can just check for that instead

AdrianGonz97:
I wonder if we could use https://github.com/Rich-Harris/esrap to make the script formatting _less_ bad? although, I'm not sure if it's compatible with our recast generated AST

manuel3108:
Finding working TS tooling is hard. That's why we are still mainly using the same tools as v1, with some exceptions. As mentioned in another issue, we might need to check if this method is potentially causing more problems than it solves!

https://github.com/svelte-add/svelte-add/blob/2f9a963fa826864c8207582b2f8fc3b83a893c44/packages/ast-tooling/index.ts#L81

benmccann:
><script>import "../app.css";
>import Header from "./Header.svelte";
>import "./styles.css";</script>

`app.css` and `styles.css` are both pretty generically named and seem indistinguishable from one another by name. Perhaps we should call it `tailwind.css` instead of `app.css`?

AdrianGonz97:
> `app.css` and `styles.css` are both pretty generically named and seem indistinguishable from one another by name. Perhaps we should call it `tailwind.css` instead of `app.css`?

The purpose of `app.css` is to be the main global stylesheet, rather than one just for tailwind imports. however, it just so happens that the demo app from `create-svelte` also includes a global stylesheet. I do prefer the name `app.css` though, rather than `styles.css`. It nicely coincides with `app.html`. also `styles.css` feels a bit redundant in name

adders/tailwindcss/config/adder.ts
```diff
@@ -92,7 +92,9 @@ export const adder = defineAdderConfig({
             content: ({ js, html }) => {
                 js.imports.addEmpty(js.ast, "../app.css");
                 const slot = html.element("slot");
```
benmccann, line 94: you could inline this inside the `if`. no need to create the element if it's not going to be used

adders/tailwindcss/config/adder.ts
```diff
@@ -92,7 +92,9 @@ export const adder = defineAdderConfig({
...
-                html.ast.childNodes.push(slot);
+                if (html.ast.childNodes.length === 0) {
```
benmccann, line 95: I imagine this would be triggered in an empty layout or script-only layout, which I suppose is fine, but I kind of expected the check for no file existing would be different than the check for empty file. It seems like we might not be able to tell those two cases apart

benmccann:
If the formatting is difficult, maybe we just need a method like `insertImport`, which adds the import after the other import lines without having to parse and serialize the whole ast?

adders/tailwindcss/config/adder.ts
AdrianGonz97: > It seems like we might not be able to tell those two cases apart

that's correct. within the context of the `content` hook, there's no easy way to tell if the file is being created or not

AdrianGonz97: > I imagine this would be triggered in an empty layout or script-only layout, which I suppose is fine

testing this now, it doesn't trigger if there's a script tag inside since that's considered a child node. If the layout is empty of text, _then_ it'll add the slot

AdrianGonz97:
> If the formatting is difficult, maybe we just need a method like insertImport, which adds the import after the other import lines without having to parse and serialize the whole ast?

hmm, I guess we _could_ change the content type for this to `text` and use `magic-string` to perform the string manipulations without having to fumble around with parsing and serializing. 

we could even provide a helper method that'll split the svelte file into sections of their respective types (markup, style, script), just like how svelte's [preprocess field](https://svelte.dev/docs/svelte-compiler#preprocess) works in the compiler

manuel3108:
Is there anything that prevents us from merging this? I think this would already be an improvement

PR #463, opened 2024-07-01
fix: alphabetize `dependencies` and `devDependencies`

### LINKED ISSUE #457: alphabetize dependencies/devDependencies
### Topic

core

### Version

2.3.6

### Expected behavior

should be alphabetized like with a newly created project

### Actual behavior

running the tailwind adder just added them to the end:
```
	"devDependencies": {
		"@fontsource/fira-mono": "^4.5.10",
		"@neoconfetti/svelte": "^1.0.0",
		"@sveltejs/adapter-auto": "^3.0.0",
		"@sveltejs/kit": "^2.0.0",
		"@sveltejs/vite-plugin-svelte": "^3.0.0",
		"svelte": "^4.2.7",
		"svelte-check": "^3.6.0",
		"tslib": "^2.4.1",
		"typescript": "^5.0.0",
		"vite": "^5.0.3",
		"tailwindcss": "^3.4.4",
		"postcss": "^8.4.38",
		"autoprefixer": "^10.4.19"
	},
```

AdrianGonz97:
closes #457

I'm not 100% sure that this is the right solution, but I'll point out my concerns below:

### DIFF
--- a/packages/ast-tooling/index.ts
+++ b/packages/ast-tooling/index.ts
@@ -176,6 +176,39 @@ export function parseJson(content: string) {
 export function serializeJson(originalInput: string, data: unknown) {
     // some of the files we need to process contain comments. The default
     // node JSON.parse fails parsing those comments.
-    // return fleece.patch(originalInput, data);
-    return fleece.stringify(data);
+    const spaces = guessIndentString(originalInput);
+    return fleece.stringify(data, { spaces });
+}
+
+// Sourced from `golden-fleece`
+// https://github.com/Rich-Harris/golden-fleece/blob/f2446f331640f325e13609ed99b74b6a45e755c2/src/patch.ts#L302
+function guessIndentString(str: string): number | undefined {
+    const lines = str.split("\n");
+
+    let tabs = 0;
+    let spaces = 0;
+    let minSpaces = 8;
+
+    lines.forEach((line) => {
+        const match = /^(?: +|\t+)/.exec(line);
+        if (!match) return;
+
+        const whitespace = match[0];
+        if (whitespace.length === line.length) return;
+
+        if (whitespace[0] === "\t") {
+            tabs += 1;
+        } else {
+            spaces += 1;
+            if (whitespace.length > 1 && whitespace.length < minSpaces) {
+                minSpaces = whitespace.length;
+            }
+        }
+    });
+
+    if (spaces > tabs) {
+        let result = "";
+        while (minSpaces--) result += " ";
+        return result.length;
+    }
 }
--- a/packages/ast-tooling/index.ts
+++ b/packages/ast-tooling/index.ts
@@ -173,9 +173,9 @@ export function parseJson(content: string) {
     return fleece.evaluate(content);
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function serializeJson(originalInput: string, data: any) {
+export function serializeJson(originalInput: string, data: unknown) {
     // some of the files we need to process contain comments. The default
     // node JSON.parse fails parsing those comments.
-    return fleece.patch(originalInput, data);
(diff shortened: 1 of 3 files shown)


packages/ast-tooling/index.ts
```diff
@@ -173,9 +173,9 @@ export function parseJson(content: string) {
...
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function serializeJson(originalInput: string, data: any) {
+export function serializeJson(originalInput: string, data: unknown) {
     // some of the files we need to process contain comments. The default
     // node JSON.parse fails parsing those comments.
-    return fleece.patch(originalInput, data);
+    // return fleece.patch(originalInput, data);
+    return fleece.stringify(data);
```
AdrianGonz97, line 180: From the docs for [`silver-fleece`](https://github.com/evertheylen/silver-fleece/blob/master/README.md), it looks like `patch` is used to preserve the formatting, but it's also interfering with our attempts of alphabetizing the dependencies. 

For instance, when we use the tailwind adder with the typography plugin included, all of the newly added deps are still be appended to the end of the dependency map:
img

packages/ast-tooling/index.ts
```diff
@@ -173,9 +173,42 @@ export function parseJson(content: string) {
...
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function serializeJson(originalInput: string, data: any) {
+export function serializeJson(originalInput: string, data: unknown) {
     // some of the files we need to process contain comments. The default
     // node JSON.parse fails parsing those comments.
-    return fleece.patch(originalInput, data);
+    const spaces = guessIndentString(originalInput);
+    return fleece.stringify(data, { spaces });
```
benmccann, line 180: we should only pass `spaces` if the file is already using spaces and not if the file is using tabs

benmccann: oh, nevermind. I see `guessIndentString` returns `undefined` for tabs, so it looks like we're okay there

benmccann (approved):

packages/ast-tooling/index.ts
manuel3108, line 180: This brings up a completely different topic that we had previously, but that does not apply at this point. 

Let's assume we have a `tsconfig.json` like this (created by `create-svelte`)
```json
{
	"extends": "./.svelte-kit/tsconfig.json",
	"compilerOptions": {
		"allowJs": true,
		"checkJs": true,
		"esModuleInterop": true,
		"forceConsistentCasingInFileNames": true,
		"resolveJsonModule": true,
		"skipLibCheck": true,
		"sourceMap": true,
		"strict": true,
		"moduleResolution": "bundler"
	}
	// Path aliases are handled by https://kit.svelte.dev/docs/configuration#alias
	// except $lib which is handled by https://kit.svelte.dev/docs/configuration#files
	//
	// If you want to overwrite includes/excludes, make sure to copy over the relevant includes/excludes
	// from the referenced tsconfig.json - TypeScript does not merge them in
}
```

With the approach we would now implement the comment would disappear, because the comments are not part of the JSON object. Currently we don't have that case active anymore, because we removed the routify? adder recently where this usecase came from. 

In order to get the best out of both worlds we should maybe introduce a new file type `json-comments` here 

https://github.com/svelte-add/svelte-add/blob/2f9a963fa826864c8207582b2f8fc3b83a893c44/packages/core/files/processors.ts#L56

and introduce some different handling, as I'm sure such a thing will happen in the future. On the other hand we could also leave it open for the future.

manuel3108 (changes requested):

packages/ast-tooling/index.ts
manuel3108: We will just need to remember that in the future! But ok for now.

manuel3108 (approved):

PR #466, opened 2024-07-02
feat: allow adders to run after another adder

### LINKED ISSUE #411: Decide how to have DB and auth adders interact with each other

manuel3108:
Closes #411 

In contrary to my expectations and what I wrote in #411 I found a pretty neat code snippet that orders the adders as expected. The solution is based on https://stackoverflow.com/a/72030336/16075084 as noted in the comment.

I have added some test data to make it easy for you to test the changes. Those changes should be removed before merging this, as they currently have no meaning. This feature will get meaning once we implement prettier and eslint adders.

| Adder | runsAfter |
|--------|--------|
| bootstrap | tailwindcss |
| bulma | tailwindcss, bootstrap |
| tailwindcss | mdsvex |
| mdsvex | - | 

You can currently execute this command to test this:
```shell
pnpx svelte-add bootstrap bulma tailwindcss mdsvex --path ./temp --default --skip-install --skip-preconditions
```

This will print the generated order the adders will be applied for the moment. In this case it generates the following output, which is the only acceptable order.

```js
[ 'mdsvex', 'tailwindcss', 'bootstrap', 'bulma' ]
```

### DIFF
--- a/.changeset/forty-tools-learn.md
+++ b/.changeset/forty-tools-learn.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/core": minor
+---
+
+feat: allow adders to run after another adder
--- a/adders/bootstrap/config/adder.ts
+++ b/adders/bootstrap/config/adder.ts
@@ -16,6 +16,7 @@ export const adder = defineAdderConfig({
     },
     options,
     integrationType: "inline",
+    runsAfter: ["tailwindcss"],
     packages: [
         { name: "bootstrap", version: "^5.3.3", dev: true },
         { name: "@popperjs/core", version: "^2.11.8", dev: true, condition: ({ options }) => options.addJavaScript },
--- a/adders/bootstrap/config/adder.ts
+++ b/adders/bootstrap/config/adder.ts
@@ -16,7 +16,6 @@ export const adder = defineAdderConfig({
     },
     options,
     integrationType: "inline",
-    runsAfter: ["tailwindcss"],
     packages: [
         { name: "bootstrap", version: "^5.3.3", dev: true },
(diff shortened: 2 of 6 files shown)


AdrianGonz97 (approved):
ahh, this is amazing!! glad you found a simple solution for it!

rnbokade:
I think, instead of checking just by the name we can also add a check on "one of" and "all of" a particular group of a adders. E.g. Future adders could include Postgres, Mysql, Turso etc. which could be part of DB group. Then There could be a group for ORMs which would contain Drizzle, Kysely, TypeORM, Prisma etc. any of which runs after any one of the DB group. Similarly there could be another group called AuthPackages which contain Lucia, AuthJS etc. both of which run after any of the ORMs.
We should also make suggestion while executing an adder directly to try to first install the dependent adder.
e.g. directly doing svelte-add Lucia should prompt user "This added should run after an adder from ORM group. Do you want to add them first." clicking yes should result in installation of that adder first. Similarly the ORM group adder would depend on DB group adder . So That again resolves that dependency by prompting user if they want to do it. 

Orrr..... we could keep the adders as they are right now... i.e. independent of each other. We write all the adders independently and then let's create a new svelte-scaffold project which uses adders behind the scene for this whole recursive dependency resolution logic.

PR #467, opened 2024-07-02
fix: preserve formatting for JS files

### LINKED ISSUE #464: tabs changed to spaces
### Topic

mdsvex

### Version

2.3.2

### Expected behavior

leave the tabs in place

it looks like there's a utility function we can use for this now: https://github.com/svelte-add/svelte-add/pull/463#discussion_r1661634219

### Actual behavior

the mdsvex adder changes all of the tabs to spaces in `svelte.config.js`

AdrianGonz97:
fixes #464

locations need to be stripped in order for node comparisons to work properly, see: https://github.com/svelte-add/svelte-add/pull/379#discussion_r1638361267

rather than stripping the ast locations from the original tree, we'll now create a clone of the tree and strip it from there during comparisons.

---

notable observation: it doesn't seem to be perfect either, but I'm not sure why. 

for instance, here's the before:
img

and here's the after:
img

however, if I format it using tabs and run the adder again, the tabs are entirely preserved... so I'm not quite sure what's going on there 😅

### DIFF
--- a/packages/ast-manipulation/package.json
+++ b/packages/ast-manipulation/package.json
@@ -13,7 +13,8 @@
         "build"
     ],
     "dependencies": {
-        "@svelte-add/ast-tooling": "workspace:^"
+        "@svelte-add/ast-tooling": "workspace:^",
+        "decircular": "^1.0.0"
     },
     "bugs": "https://github.com/svelte-add/svelte-add/issues",
     "repository": {
--- a/packages/ast-manipulation/package.json
+++ b/packages/ast-manipulation/package.json
@@ -13,7 +13,9 @@
         "build"
     ],
     "dependencies": {
-        "@svelte-add/ast-tooling": "workspace:^",
+        "@svelte-add/ast-tooling": "workspace:^"
+    },
+    "devDependencies": {
         "decircular": "^1.0.0"
     },
     "bugs": "https://github.com/svelte-add/svelte-add/issues",
--- a/.changeset/slow-chairs-tell.md
+++ b/.changeset/slow-chairs-tell.md
@@ -1,0 +1,6 @@
+---
+"@svelte-add/ast-manipulation": patch
+"@svelte-add/ast-tooling": patch
+---
+
+fix: preserve formatting for JS files
--- a/adders/mdsvex/config/adder.ts
+++ b/adders/mdsvex/config/adder.ts
@@ -20,7 +20,7 @@ export const adder = defineAdderConfig({
         {
             name: () => `svelte.config.js`,
             contentType: "script",
-            content: ({ ast, array, object, common, functions, imports, exports }) => {
+            content: ({ ast, array, object, functions, imports, exports }) => {
                 imports.addNamed(ast, "mdsvex", { mdsvex: "mdsvex" });
 
                 const { value: exportDefault } = exports.defaultExport(ast, object.createEmpty());
@@ -41,10 +41,8 @@ export const adder = defineAdderConfig({
 
                 // extensions
                 const extensionsArray = object.property(exportDefault, "extensions", array.createEmpty());
-                const svelteString = common.createLiteral(".svelte");
-                const svxString = common.createLiteral(".svx");
-                array.push(extensionsArray, svelteString);
-                array.push(extensionsArray, svxString);
+                array.push(extensionsArray, ".svelte");
+                array.push(extensionsArray, ".svx");
             },
         },
     ],
--- a/packages/ast-manipulation/js/common.ts
+++ b/packages/ast-manipulation/js/common.ts
@@ -1,4 +1,5 @@
-import { type AstKinds, type AstTypes, Walker, parseScript, serializeScript } from "@svelte-add/ast-tooling";
+import { type AstKinds, type AstTypes, Walker, parseScript, serializeScript, stripAst } from "@svelte-add/ast-tooling";
+import decircular from "decircular";
 
 export function addJsDocTypeComment(node: AstTypes.Node, type: string) {
     const comment: AstTypes.CommentBlock = {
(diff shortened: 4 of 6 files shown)


packages/ast-manipulation/package.json
```diff
@@ -13,7 +13,8 @@
         "build"
     ],
     "dependencies": {
-        "@svelte-add/ast-tooling": "workspace:^"
+        "@svelte-add/ast-tooling": "workspace:^",
+        "decircular": "^1.0.0"
```
manuel3108, line 17: Shouldn't that be a devDep?

AdrianGonz97: oops! nice catch, fixed

benmccann:
cool! but yeah, we should probably see if we can get it to be all tabs vs a mix of spaces and tabs

AdrianGonz97:
> cool! but yeah, we should probably see if we can get it to be all tabs vs a mix of spaces and tabs

playing with it a bit, it looks like if there's no trailing commas (which is how `create-svelte` applies it), it goes rogue and starts using a mix n' match 🤔

AdrianGonz97:
ehhh, I'm not so sure about this PR anymore. I can't get `recastPrint` to comply at all, even if I force it to use tabs, it still inserts spaces when updating existing files:
```js
/** @type {import('@sveltejs/kit').Config} */
const config = {
    // Consult https://kit.svelte.dev/docs/integrations#preprocessors
	// for more information about preprocessors
	preprocess: [vitePreprocess(), mdsvex()],

    kit: {
		// adapter-auto only supports some environments, see https://kit.svelte.dev/docs/adapter-auto for a list.
		// If your environment is not supported, or you settled on a specific environment, switch out the adapter.
		// See https://kit.svelte.dev/docs/adapters for more information about adapters.
		adapter: adapter()
	},

    extensions: [".svelte", ".svx"]
};

export default config;
```

Additionally, if we take into account the formatting for other adders, bringing back the location field makes it even worse for newly created files. 

consider the following case:
```
pnpm dlx https://pkg.pr.new/svelte-add/svelte-add@467 drizzle tailwindcss --path ./temp
```
if you take a peek at the generated ts modules from this PR, it's pretty bad:
```ts
// db/index.ts
import { drizzle } from "drizzle-orm/better-sqlite3";
import Database from "better-sqlite3";
import { env } from "$env/dynamic/private";
if (!env.DATABASE_URL) throw new Error("DATABASE_URL is not set");
const client = new Database(env.DATABASE_URL);
export const db = drizzle(client);


// schema.ts
import { sqliteTable, text, integer } from "drizzle-orm/sqlite-core";

export const user = sqliteTable('user', {
                        id: integer('id').primaryKey(),
                        name: text('name').notNull(),
                        age: integer('age')
                    });


// drizzle.config.ts
import { defineConfig } from "drizzle-kit";
if (!process.env.DATABASE_URL) throw new Error('DATABASE_URL is not set');

export default defineConfig({
    schema: './src/lib/server/db/schema.ts',
    dialect: 'sqlite', 
    
    dbCredentials: {
        url: process.env.DATABASE_URL,
        
    },
    verbose: true,
    strict: true
});
```

here's how they currently appear in `main`:
```
pnpm dlx svelte-add@latest drizzle tailwindcss --path ./temp
```
```ts
// db/index.ts
import { drizzle } from "drizzle-orm/better-sqlite3";
import Database from "better-sqlite3";
import { env } from "$env/dynamic/private";

if (!env.DATABASE_URL)
    throw new Error("DATABASE_URL is not set");

const client = new Database(env.DATABASE_URL);
export const db = drizzle(client);


// schema.ts
import { sqliteTable, text, integer } from "drizzle-orm/sqlite-core";

export const user = sqliteTable("user", {
    id: integer("id").primaryKey(),
    name: text("name").notNull(),
    age: integer("age")
});


// drizzle.config.ts
import { pathToFileURL } from "node:url";
import { defineConfig } from "drizzle-kit";

if (!process.env.DATABASE_URL)
    throw new Error("DATABASE_URL is not set");

export default defineConfig({
    schema: "./src/lib/server/db/schema.ts",
    dialect: "sqlite",

    dbCredentials: {
        url: process.env.DATABASE_URL
    },

    verbose: true,
    strict: true
});
```

I _could_ try to account for both by stripping the `loc` field for files that don't already exist, but I'm not sure if the added complexity is really worth it, especially when keeping the `loc` field doesn't even fully work. 

I feel like if users care about the formatting, they should really be using a formatter

manuel3108:
But it must have something to do with our tooling. I made a quick test-script and tried around with the three different files you provided, and alle the tabs / spaces and indentation is preserved as it was before, even if I decided to use some unexpected indentations:

image


Here is the test script:

```js
#!/usr/bin/env node

import { parseScript, serializeScript } from "@svelte-add/ast-tooling";
import { readFile, writeFile } from "fs/promises";

async function stuff() {
    const inFile = "test-in.js";
    const outFile = "test-out.js";

    const contentBuffer = await readFile(inFile);
    let content = contentBuffer.toString();

    const ast = parseScript(content);
    content = serializeScript(ast);

    await writeFile(outFile, content);
}

stuff();
```

AdrianGonz97:
> But it must have something to do with our tooling. I made a quick test-script and tried around with the three different files you provided, and alle the tabs / spaces and indentation is preserved as it was before, even if I decided to use some unexpected indentations:

The indentation issues I was seeing with injected nodes was due to the indentation of those created nodes when parsing from a string to an AST node. The `loc` property was preserving the extra indents.

For example:
```ts
	userSchemaExpression = common.expressionFromString(`pgTable('user', {
		id: serial('id').primaryKey(),
		name: text('name').notNull(),
		age: integer('age'),
	})`);
```

However, using `dedent` before parsing fixed it in a separate PR #470, so that's no longer a problem.

---

The first part of my post regarding injecting property nodes into a preexisting objects is still occurring, where recast's printer refuses to respect the existing indentation. I'll keep messing with it for a bit longer, but the current iteration might be the best we can do for now. 

We should also keep in mind that this is only an issue for users who _don't have Prettier installed_, which a tiny fraction. I'd much rather spend energy elsewhere if we're likely going to replace the recast printer anyway with an acorn alternative

manuel3108 (approved):

manuel3108:
LGTM

PR #470, opened 2024-07-02
fix: improve robustness of `drizzle` adder

AdrianGonz97:
While working on #467, I noticed that the robustness of the drizzle adder's config modifications could be improved. Running the adder again and changing the DB type wouldn't result in updates to the user's `drizzle.config.ts` (i.e. updating the `driver` field from `sqlite` to `mysql`). Now, if the adder is ran again with a different DB selected, the drizzle config will update accordingly.

### DIFF
--- a/adders/drizzle/config/adder.ts
+++ b/adders/drizzle/config/adder.ts
@@ -203,12 +203,16 @@ export const adder = defineAdderConfig({
                     }),
                     verbose: { type: "BooleanLiteral", value: true },
                     strict: { type: "BooleanLiteral", value: true },
+                    driver,
                 });
 
                 object.overrideProperties(objExpression, {
                     dialect: common.createLiteral(options.database),
-                    driver: driver,
                 });
+
+                // The `driver` property is only required for _some_ sqlite DBs.
+                // We'll need to remove it if it's anything but sqlite
+                if (options.database !== "sqlite") object.removeProperty(objExpression, "driver");
             },
         },
         {
--- a/adders/drizzle/config/adder.ts
+++ b/adders/drizzle/config/adder.ts
@@ -191,25 +191,21 @@ export const adder = defineAdderConfig({
                 const objExpression = exportDefault.arguments?.[0];
                 if (!objExpression || objExpression.type !== "ObjectExpression") return;
 
-                const schemaPath = common.createLiteral(`./src/lib/server/db/schema.${typescript.installed ? "ts" : "js"}`);
-                object.overrideProperty(objExpression, "schema", schemaPath);
-                object.overrideProperty(objExpression, "dialect", common.createLiteral(options.database));
+                const driver = options.sqlite === "turso" ? common.createLiteral("turso") : undefined;
+                const authToken =
+                    options.sqlite === "turso" ? common.expressionFromString("process.env.DATABASE_AUTH_TOKEN") : undefined;
 
-                const dbCredentials = object.createEmpty();
-                object.overrideProperty(dbCredentials, "url", common.expressionFromString("process.env.DATABASE_URL"));
-
-                if (options.sqlite === "turso") {
-                    object.overrideProperty(objExpression, "driver", common.createLiteral("turso"));
-                    object.overrideProperty(
-                        dbCredentials,
-                        "authToken",
-                        common.expressionFromString("process.env.DATABASE_AUTH_TOKEN"),
-                    );
-                }
-
-                object.overrideProperty(objExpression, "dbCredentials", dbCredentials);
-                object.overrideProperty(objExpression, "verbose", { type: "BooleanLiteral", value: true });
-                object.overrideProperty(objExpression, "strict", { type: "BooleanLiteral", value: true });
+                object.overrideProperties(objExpression, {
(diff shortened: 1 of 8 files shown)


adders/drizzle/config/adder.ts
```diff
@@ -222,7 +224,7 @@ export const adder = defineAdderConfig({
                         integer: "integer",
                     });
 
-                    userSchemaExpression = common.expressionFromString(`sqliteTable('user', {
+                    userSchemaExpression = common.expressionFromString(dedent`sqliteTable('user', {
```
benmccann, line 227: could we move the dedent to occur inside `expressionFromString`? or is there ever a time we wouldn't want to do that?

AdrianGonz97: heh, this is actually one of the changes that I have uncommitted and pending for #467 and it broke everything for unrelated reasons 😆 I can definitely move it here

adders/drizzle/config/adder.ts
```diff
@@ -176,38 +176,40 @@ export const adder = defineAdderConfig({
...
-                exports.defaultExport(ast, defaultExport);
+                const fallback = common.expressionFromString("defineConfig({})");
+                const { value: exportDefault } = exports.defaultExport(ast, fallback);
+                if (exportDefault.type !== "CallExpression") return;
+
+                const objExpression = exportDefault.arguments?.[0];
+                if (!objExpression || objExpression.type !== "ObjectExpression") return;
+
+                const schemaPath = common.createLiteral(`./src/lib/server/db/schema.${typescript.installed ? "ts" : "js"}`);
+                object.overrideProperty(objExpression, "schema", schemaPath);
```
benmccann, line 195: this could probably be simplified a lot if there were a method for overriding multiple properties at once:
```
object.overrideProperties(objExpression, {
    "schema": schemaPath,
    "dialect": common.createLiteral(options.database),
    "verbose": { type: "BooleanLiteral", value: true },
    "strict": { type: "BooleanLiteral", value: true },
    "dbCredentials": object.create({
        "url": common.expressionFromString("process.env.DATABASE_URL"),
        options.sqlite === "turso" ? ...{
            "driver": common.createLiteral("turso"),
            "authToken": common.expressionFromString("process.env.DATABASE_AUTH_TOKEN")
        }: ...{}
    })
})
```

benmccann: I wonder if we should really be overriding though? If they've already got a file there, I'm not sure they'd want us to mess with it. It seems simpler and safer to just create a single string and write it out if the file doesn't exist already

AdrianGonz97: hmm, perhaps we could just override the important parts then? those would be: `dialect` and `driver`. everything else could remain untouched

AdrianGonz97: > I wonder if we should really be overriding though?

i guess it depends on the ethos of `svelte-add`. are we a "run us once and you're done" app, or can it be ran again for it to be repaired/updated/"add more stuff" (in the case of tailwind plugins)

benmccann: > hmm, perhaps we could just override the important parts then? those would be: dialect and driver. everything else could remain untouched

what happens if you have a postgres dialect and driver, but sqlite connection string and schema?

> i do love the idea of this utility though. will implement it shortly

thinking about it more, I wonder if you could do something even easier to use like this:
```
parse(`{
    "schema": ${JSON.stringify(schemaPath)},
    "dialect": ${JSON.stringify(options.database)},
    "verbose": true,
    "strict": true,
    "dbCredentials": {
        "url": common.expressionFromString("process.env.DATABASE_URL"),
        ${options.sqlite === "turso" ? 
            `"driver": "turso",
            "authToken": process.env.DATABASE_AUTH_TOKEN` : ''
        }
    }
}`)
```

> i guess it depends on the ethos of svelte-add. are we a "run us once and you're done" app, or can it be ran again for it to be repaired/updated/"add more stuff" (in the case of tailwind plugins)

we could skip updates by default and just opt-in for specific things like additional plugins where it's more adding stuff vs updating stuff

AdrianGonz97: > thinking about it more, I wonder if you could do something even easier to use like this:
> 
> ```
> parse(`{
>     "schema": ${JSON.stringify(schemaPath)},
>     "dialect": ${JSON.stringify(options.database)},
>     "verbose": true,
>     "strict": true,
>     "dbCredentials": {
>         "url": common.expressionFromString("process.env.DATABASE_URL"),
>         ${options.sqlite === "turso" ? 
>             `"driver": "turso",
>             "authToken": process.env.DATABASE_AUTH_TOKEN` : ''
>         }
>     }
> }`)
> ```

this is actually quite similar to how i originally had it before this PR 😆:

https://github.com/svelte-add/svelte-add/blob/2c75392e27bd5c47d40df16489c660cf4625b389/adders/drizzle/config/adder.ts#L196-L210

i opted to replace it because i wanted the individual properties to be a bit more dynamic and overridable rather than having it be an all or nothing change. i think what you first proposed is an excellent solution that feels very ergonomic!

manuel3108: I like the now chosen approach and think it's the best fit for the future. resolving this.

Issue #461, opened 2024-07-01
tests with fixtures


benmccann:
### Topic

all

### Description

The tailwind adder is broken right now (https://github.com/svelte-add/svelte-add/issues/459). It'd be nice if we had tests that could catch that before a PR gets merged

### Describe the solution you'd like

Use vitest fixtures. Check that the generated project matches what is expected. That will also help us in code review because we can look at the generated project and see what we think about the diff. We can also setup a script in the package.json to automatically update the fixtures so that they're not hard to maintain when we make changes to the code

benmccann:
oh, I just noticed that this is a bit of a duplicate of https://github.com/svelte-add/svelte-add/issues/356

manuel3108:
I don't think that it really is. The other issue would just make sure that `Welcome to Sveltkit` (or whatever) is rendered. But this would be way better for code review and so one. 

The only thing that makes me re-consider this is making `vitest` work with the adders. I tried this in the past, and failed. But we might have a better chance now that #446 is implemented.

manuel3108:
Is it possible that you mean [`snapshots`](https://vitest.dev/guide/snapshot.html)? I could not find much about [`fixtures`](https://vitest.dev/guide/test-context.html#fixture-initialization) in their docs.

PR #480, opened 2024-07-05
chore: use vitest

manuel3108:
Relates / in preparation for #461 

Command to test adders locally will slightly change with this:
```
pnpm test tailwindcss
```

will now be
```
pnpm test -t tailwindcss
```

You can also use `pnpm test:ui` which will also spin up vitest ui, although the ui is pretty barebones and does not allow running filtered tests in example.

Currently a draft, as I needed to switch from `pnpm` to `npm` for the testing in order to make everything work. Now trying to figure out what causes the problem


manuel3108:
For whatever reason this works in CI.....

benmccann:
I believe @dominikg setup the tests for `create-svelte` and handled whatever pnpm magic was needed there to make things work, so perhaps he'd be able to lend some advice on this one

manuel3108:
As soon as I switch this line from using `npm` to `pnpm` 26 out of 51 one tests start failing on windows:
https://github.com/svelte-add/svelte-add/blob/0bd41a87989daa52e7123ee0cca46de1928d7631/packages/testing-library/utils/workspace.ts#L18

I tried to reproduce the error in a codespace, but everything seems to work there, as in CI (well expect for the storybook tests, but I'm not interested in these atm). Also tried a to do a fresh clone on my machine, but I only got the same error as before.

dominikg:
> I believe @dominikg setup the tests for `create-svelte` and handled whatever pnpm magic was needed there to make things work, so perhaps he'd be able to lend some advice on this one

tests for create-svelte are special as you have to prevent the local monorepo packages from bleeding in. so they generate a separate monorepo root in a temp dir. its weird and pretty specific to create-svelte.

a similar pattern could be used for svelte-add but not sure you need it unless you want e2e-tests that validate the adder worked by installing and building the e2e project again. but I'd be fine with snapshots validating expected file changes

manuel3108:
Closing in favor of #544

Issue #489, opened 2024-07-15
[Prettier Adder]: E404 npm ERR! 404 Not Found


shyakadavis:
### Topic

cli

### Version

N/A

### Expected behavior

When running `pnpm dlx @svelte-add/prettier@latest` or `npx @svelte-add/prettier@latest`, I expected `prettier` addeed to my project.

### Actual behavior

Instead, I am met with these errors:

With `pnpm dlx @svelte-add/prettier@latest`
```
/Users/davis/Library/Caches/pnpm/dlx/4zjxuv3jzwytpnyk2ixns5fe5m/190b7deb780-12873:
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/@svelte-add%2Fprettier: Not Found - 404

This error happened while installing a direct dependency of /Users/davis/Library/Caches/pnpm/dlx/4zjxuv3jzwytpnyk2ixns5fe5m/190b7deb780-12873

@svelte-add/prettier is not in the npm registry, or you have no permission to fetch it.

An authorization header was used: Bearer npm_[hidden]
```

With `npx @svelte-add/prettier@latest`:
```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@svelte-add%2fprettier - Not found
npm ERR! 404
npm ERR! 404  '@svelte-add/prettier@latest' is not in this registry.
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! A complete log of this run can be found in: /Users/davis/.npm/_logs/2024-07-15T19_31_09_281Z-debug-0.log
```

manuel3108:
This syntax is no longer supported. Please consider using 

`pnpm dlx svelte-add prettier`

shyakadavis:
Hey, @manuel3108 

Ah, I see. Thanks. Was misled since I was coming from the docs/website.

I believe a change to the docs is really needed, otherwise someone else might be misled.

Happy to submit a P.R if you're up to it. 🙂

PR #490, opened 2024-07-16
docs: add correct adder script

shyakadavis:
Follow-up on #489 

Right now, when I click an adder option from the docs (`+`), it configures the old syntax of `@svelte-add/{something}`

I didn't test locally, unfortunately, because the local dev server threw an error:

```
[plugin:vite:import-analysis] Failed to resolve entry for package "@svelte-add/config". The package may have incorrect main/module/exports specified in its package.json.
``` 

So please feel free to close if this is wrong.

### DIFF
--- a/.changeset/cyan-kings-call.md
+++ b/.changeset/cyan-kings-call.md
@@ -1,0 +1,5 @@
+---
+"@svelte-add/website": patch
+---
+
+use correct command if only selecting one adder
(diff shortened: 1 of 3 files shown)


manuel3108:
Thanks for your PR!

We already updated the commands in the readmes, but we apparently forgot to update the website, so thank you for catching that. The code you supplied was generating `npx svelte-add bulma@latest` which is incorrect, because latest was at the wrong place. It should be placed directly after `svelte-add`. I have already fixed that for you.

As for the error you were reporting: You are absolutely right, this is what happens if we don't have clear documentation on what do to before starting to develop on the website. I have also added the mentioned docs in my last commit, see [here](https://github.com/shyakadavis/svelte-add/blob/docs/fix-adder-script/CONTRIBUTING.md#website)

PR #493, opened 2024-07-19
chore: workspace deps should specify exact versions on publish

### LINKED ISSUE #457: alphabetize dependencies/devDependencies

AdrianGonz97:
At the moment, trying to run older versions of `svelte-add` is kind of broken.

For instance, running:
```bash
pnpx svelte-add@2.3.0
```
will break due to `prettier` being added to the config when it doesn't exist locally in this package yet (it was added in `2.3.7`). I suspect this is partially the reason why #457 is broken in some cases. 

This is happening because range specifiers are being using when our workspace packages are referencing each other with `workspace:^`. Instead, we should be using `workspace:*` to specify exact versions on publish (this would effectively be equivalent to us bundling everything into a single package, but due to build times with AST tooling, it's not viable).

When publishing with `workspace:^`, workspace deps will have their versions set to `^2.3.5`. With `workspace:*`, the version will be set to `2.3.5` instead.

Also adding a changeset for `ast-tooling` to fix #457

### DIFF
--- a/.changeset/two-avocados-brush.md
+++ b/.changeset/two-avocados-brush.md
@@ -1,5 +1,6 @@
 ---
 "@svelte-add/ast-manipulation": patch
+"@svelte-add/testing-library": patch
 "@svelte-add/website": patch
 "@svelte-add/core": patch
 "svelte-add": patch
--- a/.changeset/two-avocados-brush.md
+++ b/.changeset/two-avocados-brush.md
@@ -1,0 +1,9 @@
+---
+"@svelte-add/ast-manipulation": patch
+"@svelte-add/testing-library": patch
+"@svelte-add/website": patch
+"@svelte-add/core": patch
+"svelte-add": patch
+---
+
+fix: pin workspace dependencies versions
--- a/.changeset/two-avocados-brush.md
+++ b/.changeset/two-avocados-brush.md
@@ -1,6 +1,5 @@
 ---
 "@svelte-add/ast-manipulation": patch
-"@svelte-add/testing-library": patch
 "@svelte-add/website": patch
 "@svelte-add/core": patch
 "svelte-add": patch
--- a/packages/dev-utils/package.json
+++ b/packages/dev-utils/package.json
@@ -6,9 +6,9 @@
     "license": "MIT",
     "bin": "./index.js",
     "devDependencies": {
-        "@svelte-add/ast-manipulation": "workspace:^",
-        "@svelte-add/ast-tooling": "workspace:^",
-        "@svelte-add/core": "workspace:^"
+        "@svelte-add/ast-manipulation": "workspace:*",
+        "@svelte-add/ast-tooling": "workspace:*",
+        "@svelte-add/core": "workspace:*"
     },
     "dependencies": {
         "npm-check-updates": "^16.14.20"
--- a/packages/testing-library/package.json
+++ b/packages/testing-library/package.json
@@ -13,7 +13,7 @@
         "build"
     ],
     "dependencies": {
-        "@svelte-add/core": "workspace:*",
+        "@svelte-add/core": "workspace:^",
         "playwright": "^1.44.1",
         "create-svelte": "^6.3.0"
     },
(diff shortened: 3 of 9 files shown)


packages/testing-library/package.json
```diff
@@ -13,7 +13,7 @@
         "build"
     ],
     "dependencies": {
-        "@svelte-add/core": "workspace:^",
+        "@svelte-add/core": "workspace:*",
```
benmccann, line 16: Maybe this one makes sense to leave as `^`?

.changeset/two-avocados-brush.md
```diff
@@ -0,0 +1,9 @@
+---
+"@svelte-add/ast-manipulation": patch
+"@svelte-add/testing-library": patch
```
benmccann, line 3:
```suggestion
```

packages/website/package.json
```diff
@@ -24,8 +24,8 @@
         "vite": "^5.3.0"
     },
     "dependencies": {
-        "@svelte-add/adders": "workspace:^",
-        "@svelte-add/config": "workspace:^",
-        "@svelte-add/core": "workspace:^"
+        "@svelte-add/adders": "workspace:*",
```
benmccann, line 27: Should these remain as `^`? I'm not as familiar with the website package

packages/dev-utils/package.json
```diff
@@ -6,9 +6,9 @@
     "license": "MIT",
     "bin": "./index.js",
     "devDependencies": {
-        "@svelte-add/ast-manipulation": "workspace:^",
-        "@svelte-add/ast-tooling": "workspace:^",
-        "@svelte-add/core": "workspace:^"
+        "@svelte-add/ast-manipulation": "workspace:*",
```
benmccann, line 9: I wonder if these should be left as `^` too since they're not being distributed to users

packages/website/package.json
AdrianGonz97: For the website, it doesn't really matter since they'll always be using the latest in `main`, so either method is fine

packages/dev-utils/package.json
AdrianGonz97: Same thing as the website, it doesn't really matter which one it uses since it's not published, so it'll always use whatever versions in the current workspace

packages/website/package.json
AdrianGonz97: Giving it some more thought, I think a good litmus test for this would be to ask: "should these be bundled?". If so, then we should leave it with a `workspace:*`. If not, then we should either consider using `workspace:^`, or perhaps making the dep a `peerDependency` _and_ use `workspace:^` (I'm thinking perhaps `testing-library` should be a peer-dep instead).

benmccann: My worry was that changesets might think versions need to be bumped more often when using `*`. I don't know if that's really something that will be an issue, but thought perhaps `^` could be safer from that respect. Otherwise I agree they should be equivalent when used in these packages that aren't published

AdrianGonz97: AFAIK, changesets isn't really concerned with how specific versions of packages relate to each other. All it cares about is who is a dependent of who, so the range specifier really doesn't matter in that regard

manuel3108: Yeah, changesets shouldn't care about this change.

manuel3108:
What a catch! Thanks so much.

Issue #472, opened 2024-07-03
Script for adder generation.


rnbokade:
### Topic

Base

### Description

To accelerate the development and for maintaining standards in boilerplate while adding new adders, can we create a CLI to that would be available only within the repository context, allowing us to add new adder package by a cli wizard, which asks name of adder, dependencies, options etc. within the wizard, resulting in quicker scaffolding of a new adder. This could greatly help speed up development of community adders, while maintaining the standards.

manuel3108:
There is still a lot of work to do regarding automation and community adders. Community adders are planned, but are not even close to be implemented anytime soon, that's why documentation is lacking. 
I'm not sure if a cli is the correct way do this as it would generate maintenace overhead, but we should at least have appropriate templates once we get there!

ieedan:
I am working on a package right now [template-factory](https://github.com/ieedan/template-factory-js) that could be useful for creating a CLI if that was the route you wanted to go. I think it would probably limit the overhead you are talking about if you wanted to have an internal CLI.

### How it would work
You could create a js/ts script in the root of the project. Add a script in package.json like `templates:create` or something.

You would have a `template` under the adders directory that you would reference from the script. 

You can run any other code or prompts to help make steps easier such as allowing them to pick a category or create a new one from the CLI.

Initially it could look something like this, this just prompts the user and copies the files to where they specify but you could add prompts and such to improve the workflow.
```ts
import { create } from 'template-factory';
import * as util from 'template-factory/util';

const main = async () => {
  await create({
    name: 'templates:create',
    version: '1.0.0',
    templates: [
      {
        name: 'custom-adder',
        // probably don't actually need the path resolver here
        path: util.relative('./adders/template', import.meta.url),
        flag: 'custom-adder',
        // you could add prompts down here to improve setup
      }
    ]
  })
}

main(); 
```

Let me know if this is something that makes and I can try and model it better in a PR! If not no worries.

manuel3108:
Thanks for bringing this to my attention again, and thanks for putting in the hard work @ieedan.

Since this issue was raised we had some restructuring going, which made generating new adders much easier. See [this commit](https://github.com/svelte-add/svelte-add/commit/8fbdd22f8e9ab7bac79f9a1de03df4373396dd12). Therefore I don't think such a cli makes sense at this point, as this would cause more Maintenace overhead than actual benefit.

Closing this.

PR #494, opened 2024-07-19
internal feat: CLI to generate new adders 

ieedan:
Draft PR for the issue #472.

Adds a CLI to generate new adders using [template-factory](https://github.com/ieedan/template-factory-js)

I thought I would just implement this to show what it might look like.

Allows anyone wanting to add an adder to run:
```bash
pnpm run adders:create adders/[YOUR ADDER]
```

From here it can take them through some basic configuration:
```
┌   adders:create  v1.0.0 
│
◇  Created analytics
│
◇  Description of the adder
│  Allows you to collect data about your application.
│
◇  Link to documentation for the adder
│  https://analytics.dev
│
◇  Is this an adder for Svelte?
│  Yes
│
◇  Is this an adder for SvelteKit?
│  Yes
│
◇  What type of integration is this?
│  inline
│
└  You're all set!
```

It would then generate the required files based on a template located in `adders/template`.

May also help with #367

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,16 +33,12 @@ pnpm changeset
 
 ## create new adder
 
--   stop development server
--   duplicate existing adder, and make some minor modification (like package name)
--   delete its `node_modules` and `build` folder
--   delete the `CHANGELOG.md` file
--   add the new adders as a peer dependency to `svelte-add`
--   set an appropriate package name & version version in `package.json`
--   run `pnpm install` (ignore the warnings)
--   add your adder to one of the categories in [`./packages/config/adders/official.ts`](./packages/config/adders/official.ts)
--   start development server `pnpm build:dev`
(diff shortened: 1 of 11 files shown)


benmccann:
It'd probably help to see an example adder that was created with this, but I have a hard time imagining how it could work or do more than what's described in https://github.com/svelte-add/svelte-add/issues/445

ieedan:
> It'd probably help to see an example adder that was created with this, but I have a hard time imagining how it could work or do more than what's described in #445

Yes it is very primitive and won't help past the initial setup of the adder. 

The main function is to copy over the structure and fill in a few of the required properties when initially setting up a new adder and wouldn't be exposed to any user of the CLI. 

It will generate the following structure for you
```
[YOUR ADDER]
├── config
│   ├── adder.ts
│   ├── checks.ts
│   ├── options.ts
│   └── tests.ts
├── index.ts
└── [YOUR ADDER].svg
```

The only modifications that are made to files currently are in the `config/adder.ts`. I think it may also be cool to add it to a category in `official.ts` by prompting for an existing category or to create a new one.

If that's not something that is worth having then that's totally fine too.

PR #492, opened 2024-07-19
fix: Replace `@tailwind` directives for `@import`

AdrianGonz97:
Addresses one of the issues from #482

Replaces the tailwind directives:
```css
@tailwind base;
@tailwind components;
@tailwind utilities;
```
for imports:
```css
@import "tailwindcss/base";
@import "tailwindcss/components";
@import "tailwindcss/utilities";
```

This was done for better compatibility when the adder is used in preexisting projects that have `@import` rules already applied. See https://github.com/svelte-add/svelte-add/issues/482#issuecomment-2239733898 for more details.

PR #496, opened 2024-07-20
fix(tailwind): ensure `@import` rules are single spaced when added

### LINKED ISSUE #482: Unnecessary new lines added
### Topic

tailwind

### Version

2.3.4

### Expected behavior

```
@tailwind base;
@tailwind components;
@tailwind utilities;

@import '@fontsource/fira-mono';

:root {
	--font-body: Arial, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu,
		Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
	--font-mono: 'Fira Mono', monospace;
	--color-bg-0: rgb(202, 216, 228);
	--color-bg-1: hsl(209, 36%, 86%);
	--color-bg-2: hsl(224, 44%, 95%);
	--color-theme-1: #ff3e00;
	--color-theme-2: #4075a6;
	--color-text: rgba(0, 0, 0, 0.7);
	--column-width: 42rem;
	--column-margin-top: 4rem;
	font-family: var(--font-body);
	color: var(--color-text);
}
```

### Actual behavior

```
@tailwind base;

@tailwind components;

@tailwind utilities;

@import '@fontsource/fira-mono';

:root {
	--font-body: Arial, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu,
		Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
	--font-mono: 'Fira Mono', monospace;
	--color-bg-0: rgb(202, 216, 228);
	--color-bg-1: hsl(209, 36%, 86%);
	--color-bg-2: hsl(224, 44%, 95%);
	--color-theme-1: #ff3e00;
	--color-theme-2: #4075a6;
	--color-text: rgba(0, 0, 0, 0.7);
	--column-width: 42rem;
	--column-margin-top: 4rem;
	font-family: var(--font-body);
	color: var(--color-text);
}
```

AdrianGonz97:
Note: this PR should be merged after #492 

Fixes #482

`@import` rules will now be added with the correct spacing:
```css
@import "tailwindcss/base";
@import "tailwindcss/components";
@import "tailwindcss/utilities";
@import "@fontsource/fira-mono";

:root {
...
```

### DIFF
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -72,10 +72,18 @@ export const adder = defineAdderConfig({
             name: () => "src/app.css",
             contentType: "css",
             content: ({ ast, addAtRule }) => {
-                const atRules = ["utilities", "components", "base"];
-                for (const name of atRules) {
-                    addAtRule(ast, "tailwind", name);
+                const atRules = ['"tailwindcss/utilities"', '"tailwindcss/components"', '"tailwindcss/base"'];
+                const firstNode = ast.first;
+                const nodes = atRules.map((name) => addAtRule(ast, "import", name));
+
+                if (firstNode !== nodes.at(-1) && firstNode?.type === "atrule" && firstNode.name === "import") {
+                    firstNode.raws.before = "\n";
                 }
+
+                nodes.forEach((n, i) => {
+                    // prevents a new line being added at the top of the stylesheet
+                    if (i !== nodes.length - 1) n.raws.before = "\n";
+                });
             },
         },
         {
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -76,7 +76,7 @@ export const adder = defineAdderConfig({
                 const firstNode = ast.first;
                 const nodes = atRules.map((name) => addAtRule(ast, "import", name));
 
-                if (firstNode !== nodes.at(-1) && firstNode?.type === "atrule" && firstNode.name === "import") {
+                if (firstNode !== ast.first && firstNode?.type === "atrule" && firstNode.name === "import") {
                     firstNode.raws.before = "\n";
                 }
 
--- a/adders/tailwindcss/config/adder.ts
+++ b/adders/tailwindcss/config/adder.ts
@@ -71,19 +71,30 @@ export const adder = defineAdderConfig({
         {
             name: () => "src/app.css",
             contentType: "css",
-            content: ({ ast, addAtRule }) => {
-                // `addAtRule` prepends the rules to the tree, so we reverse the order
-                // such that the last element ends up at the top of the stylesheet
-                const imports = ['"tailwindcss/utilities"', '"tailwindcss/components"', '"tailwindcss/base"'];
+            content: ({ ast, AtRule }) => {
+                const layers = ["base", "components", "utilities"];
                 const originalFirst = ast.first;
-                const nodes = imports.map((name) => addAtRule(ast, "import", name));
+
+                let prev: typeof originalFirst;
+                const nodes = layers.map((layer) => {
+                    const param = `"tailwindcss/${layer}"`;
+                    const found = ast.nodes.find((x) => x.type === "atrule" && x.name === "import" && x.params === param);
+
+                    if (found) return (prev = found);
+
+                    const rule = new AtRule({ name: "import", params: param });
+                    if (prev) ast.insertAfter(prev, rule);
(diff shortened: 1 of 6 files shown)


adders/tailwindcss/config/adder.ts
```diff
@@ -72,10 +72,18 @@ export const adder = defineAdderConfig({
             name: () => "src/app.css",
             contentType: "css",
             content: ({ ast, addAtRule }) => {
-                const atRules = ["utilities", "components", "base"];
-                for (const name of atRules) {
-                    addAtRule(ast, "tailwind", name);
+                const imports = ['"tailwindcss/utilities"', '"tailwindcss/components"', '"tailwindcss/base"'];
+                const firstNode = ast.first;
```
benmccann, line 76: I think something like `originalFirst` might be clearer when used later. E.g. `firstNode !== ast.first` vs `originalFirst !== ast.first`

adders/tailwindcss/config/adder.ts
```diff
@@ -72,10 +72,18 @@ export const adder = defineAdderConfig({
...
+                const nodes = imports.map((name) => addAtRule(ast, "import", name));
+
+                if (firstNode !== ast.first && firstNode?.type === "atrule" && firstNode.name === "import") {
+                    firstNode.raws.before = "\n";
                 }
+
+                nodes.forEach((n, i) => {
+                    // prevents a new line being added at the top of the stylesheet
+                    if (i !== nodes.length - 1) n.raws.before = "\n";
```
benmccann, line 85: it's not really clear to me how this works. would it otherwise be `\n\n`?

benmccann:
It's cool you figured out how to make this work. I wonder if there's an advantage to using ast manipulation vs string manipulation? It feels like this might be simpler with string manipulation

AdrianGonz97:
> I wonder if there's an advantage to using ast manipulation vs string manipulation? It feels like this might be simpler with string manipulation

I don't see any reason why it wouldn't work. I can give it a try a little later when I have another free moment

adders/tailwindcss/config/adder.ts
```diff
@@ -71,19 +71,30 @@ export const adder = defineAdderConfig({
...
                 if (originalFirst !== ast.first && originalFirst?.type === "atrule" && originalFirst.name === "import") {
                     originalFirst.raws.before = "\n";
                 }
 
-                // We remove the last node to avoid adding a newline at the top of the stylesheet
-                nodes.pop();
+                // We remove the first node to avoid adding a newline at the top of the stylesheet
+                nodes.shift();
 
                 // Each node is prefixed with single newline, ensuring the imports will always be single spaced.
```
benmccann, line 99: it looks like we're not using `addAtRule` anymore, but I wonder if this logic would make sense to have live in there or some common util function?

AdrianGonz97:
After prototyping a string manipulation implementation, I'm finding the AST version to be a bit more robust. There are a few edge cases, while unlikely, that could cause the adder to end up breaking the user's app (mainly having to do with whitespace and all the possible variations that an existing import declaration can have). 

The AST also makes it easier to repair incomplete installations as well. Consider the following input:
```css
@import "tailwindcss/base";
@import "tailwindcss/utilities";
@import "@fontsource/fira-mono";
```

output:
```css
@import "tailwindcss/base";
@import "tailwindcss/components";
@import "tailwindcss/utilities";
@import "@fontsource/fira-mono";
```

we can guarantee that the `components` import will be added in the correct location, which is pretty handy!

adders/tailwindcss/config/adder.ts
AdrianGonz97: we certainly could! probably in the form of an `addImport` method

benmccann: Could the `before = "\n"` part live there too?

AdrianGonz97: it _could_, but i'd rather have the caller have agency over those specifics for now - perhaps if another adder requires it in the future, it could be added there as well

benmccann: Double spacing seems like something you'd rarely want to do though. I imagine if you set the default to single spacing that the caller could then override as needed. There could potentially be some option to control spacing as another option - though I'm not really sure what you'd want besides single-spaced by default and then let the caller override as needed

manuel3108 (approved):

manuel3108 (approved):

PR #501, opened 2024-07-23
chore: update commands to match other Svelte projects

benmccann:


manuel3108 (dismissed):
I additionally updated the docs

package.json
```diff
@@ -9,8 +9,8 @@
         "prettier:fix": "pnpm prettier . --write",
```
AdrianGonz97, line 9: while we're at it, perhaps we can also update format and lints as well?
```suggestion
        "format": "prettier . --write",
```

benmccann: good idea!

manuel3108: Sounds smart. How would you proceed with the other commands?

```bash
"prettier:check": "pnpm prettier . --check",  # format:check
"prettier:fix": "pnpm prettier . --write", # format
"eslint:check": "pnpm eslint .", # lint
"eslint:fix": "pnpm eslint --fix .", # lint:check
```

.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
               run: pnpm types:check
```
AdrianGonz97, line 56: we can probably rename this one to just `check` as well and we should be good to go

package.json
```diff
@@ -5,12 +5,10 @@
...
     "scripts": {
         "ci:publish": "changeset publish",
-        "prettier:check": "pnpm prettier . --check",
-        "prettier:fix": "pnpm prettier . --write",
-        "eslint:check": "pnpm eslint .",
-        "eslint:fix": "pnpm eslint --fix .",
-        "build:dev": "rollup --config --watch",
-        "build:prod": "rollup -c",
+        "format": "prettier . --write && pnpm eslint --fix .",
+        "lint": "pnpm prettier . --check && pnpm eslint .",
```
AdrianGonz97, line 9:
```suggestion
        "format": "prettier . --write && eslint --fix .",
        "lint": "prettier . --check && eslint .",
```

.github/workflows/ci.yml
benmccann: done

AdrianGonz97 (approved):
awesome!!

manuel3108 (approved):

PR #502, opened 2024-07-23
feat: ESLint adder

### LINKED ISSUE #497: Port over eslint setup
### Topic

ealint

### Description

Someone just shared that the eslint setup from create-svelte doesn't setup VS Code for eslint. When we port that one over, perhaps we should have a question about configuring it for VS Code?

https://sveltejs.github.io/eslint-plugin-svelte/user-guide/#visual-studio-code

_No response_

AdrianGonz97:
TODO:
- [x] ensure this runs _after_ the prettier adder if prettier is also selected
- [x] (closes #497) if `.vscode/settings.json` is detected and `eslint.validate` exists, ensure that svelte files are included in the array
- [x] ~~figure out what to do with types. should we also add `@types/eslint__js` and `@types/eslint-config-prettier`?~~ (we'll worry about it later)

Need to fix:
- [x] if you modify any of the object literals (such as changing the order of the props) and rerun the adder, it'll readd the object literal again 
(Update: decided to ditch mutating the config if it already exists. atm, it's safer to leave it as is than it is to attempt a repair)
- [x] configs also don't go in the correct order if you run the adder in an already existing config ~~(not sure if this is feasible yet)~~
(Update: same reason as the update above)
- [x] a newline space gets added between `files` and `languageOptions` for some reason:
(Update: I yield. I can't find a way to disable this bizarre behavior. Hopefully this won't be a problem when we (eventually) switch over to acorn) 
```js
{
	files: ['**/*.svelte'],

	languageOptions: {
		parserOptions: {
			parser: ts.parser
		}
	}
},
```


adders/eslint/config/adder.ts
```diff
@@ -0,0 +1,110 @@
...
+            logo: "./eslint.svg",
+            keywords: ["eslint", "code", "linter"],
+            documentation: "https://eslint.org/",
+        },
+    },
+    options,
+    integrationType: "inline",
+    packages: [
+        { name: "eslint", version: "^9.7.0", dev: true },
+        { name: "@types/eslint", version: "^9.6.0", dev: true },
```
benmccann, line 20: lol. that was fast. I didn't expect you to have incorporated https://github.com/sveltejs/kit/pull/12496 already :smile:

AdrianGonz97: ha! that's actually just pure coincidence - I saw [this on twitter](https://x.com/geteslint/status/1815448820647411737) the other day and added it as part of the initial implementation last night 😆

benmccann:
> ensure this runs after the prettier adder if prettier is also selected

or would it be better to make them be able to be run independently? that way if the user only sets up eslint and then comes back later and runs the prettier adder it will be handled correctly

> figure out what to do with types. should we also add @types/eslint__js and @types/eslint-config-prettier?

I'm not super familiar with those, but they're not added by `create-svelte` so shouldn't be a blocker at least

> add question for setting up workspace settings for vscode

I take it back that we should have a question about this. Just added some details: https://github.com/svelte-add/svelte-add/issues/497#issuecomment-2248605197

AdrianGonz97:
> or would it be better to make them be able to be run independently? that way if the user only sets up eslint and then comes back later and runs the prettier adder it will be handled correctly

they'll definitely be able to run independently from each other. ensuring eslint runs _after_ prettier just makes it easier (and more reliable) for certain detections (such as not need to worry about finding the `svelte.configs["flat/recommended"]` that we do [here](https://github.com/svelte-add/svelte-add/blob/1d33c8ce392cc9a60716af0dc5facb2f9fea6a1a/adders/prettier/config/adder.ts#L118-L128))

> I'm not super familiar with those, but they're not added by create-svelte so shouldn't be a blocker at least

with `create-svelte` you actually do get errors for missing declaration files for these modules:

img

> I take it back that we should have a question about this. Just added some details: https://github.com/svelte-add/svelte-add/issues/497#issuecomment-2248605197

ah, makes sense. will update the comment 👍

benmccann:
> they'll definitely be able to run independently from each other. ensuring eslint runs after prettier just makes it easier (and more reliable) for certain detections (such as not need to worry about finding the svelte.configs["flat/recommended"] that we do [here](https://github.com/svelte-add/svelte-add/blob/1d33c8ce392cc9a60716af0dc5facb2f9fea6a1a/adders/prettier/config/adder.ts#L118-L128))

I was suggesting that it might be nice to be able to do the detection so that the user gets the full experience if setting up eslint followed by prettier

> with create-svelte you actually do get errors for missing declaration files for these modules

I remembered now that @hyunbinseo filed an issue for this: https://github.com/sveltejs/kit/issues/12109. it seems we were blocked by upstream. @hyunbinseo did you end up filing an issue upstream or know if this might be unblocked yet?

AdrianGonz97:
> I was suggesting that it might be nice to be able to do the detection so that the user gets the full experience if setting up eslint followed by prettier

sorry, I wasn't being clear in my reply - in reality, it really doesn't matter which order they run in. the user could run `eslint` first, followed by `prettier` (and vice versa), and everything will work just fine. all I wanted was to ensure that if they happen select both adders at the same time, then I'd _prefer_ if `prettier` ran first just so that the executed logic that lives in `eslint` is guaranteed to run since it's a _tiny bit_ more reliable (it also does a little bit less work too) than the logic that resides in the `prettier` adder

it's not really a big deal though, just a small thing

AdrianGonz97:
ah, small update on the type declaration errors:

those errors only popup if `.svelte-kit` hasn't been generated yet. since `.svelte-kit/tsconfig.json` doesn't include `eslint.config.js` in it's `include` array, it won't be type checked and the red squigglies go away lol

I wonder if we should add `// @ts-check` at the top of the file 🤔

benmccann:
> since .svelte-kit/tsconfig.json doesn't include eslint.config.js in it's include array, it won't be type checked and the red squigglies go away lol

Ah. We'll fix that in SvelteKit 3. We wanted to fix it earlier, but it was deemed a breaking change. Anyway, that gives us awhile before we really need to worry about it

AdrianGonz97:
recast's printer is so weird...

img

by default, it adds a new line above object literals if there's a property above it, wtf

not sure how to disable this behavior yet

hyunbinseo:
I haven't filed an issue. I've just updated `@types/eslint` and it fixed the issue.

> I remembered now that @hyunbinseo filed an issue for this: [sveltejs/kit#12109](https://github.com/sveltejs/kit/issues/12109). it seems we were blocked by upstream. @hyunbinseo did you end up filing an issue upstream or know if this might be unblocked yet?

I can confirm that a missing `.svelte-kit` directory causes these errors. Running `pnpm check` fixes them.

```js
// eslint.config.js
import js from '@eslint/js'; // Could not find a declaration file ~
import prettier from 'eslint-config-prettier'; // Could not find a declaration file ~
```

adders/eslint/config/adder.ts
```diff
@@ -0,0 +1,134 @@
...
+	],
+	files: [
+		{
+			name: () => 'package.json',
+			contentType: 'json',
+			content: ({ data }) => {
+				data.scripts ??= {};
+				const scripts: Record<string, string> = data.scripts;
+				const LINT_CMD = 'eslint .';
+				scripts['lint'] ??= LINT_CMD;
```
manuel3108, line 47: Should we also add `eslint --fix .` to the `format` command, as in our repo?

manuel3108 (dismissed):

adders/eslint/config/adder.ts
AdrianGonz97: we could, but I was basing this off of the existing template that's generated from `create-svelte` with `prettier+eslint` which doesn't include that. I'm leaning towards keeping it as-is and leaving it out

manuel3108: Yeah, didn't think about that. I think that aligns with what we were doing with the prettier adder.

manuel3108 (approved):

PR #505, opened 2024-07-24
chore: adjust prettier config

manuel3108:
As discussed yesterday, we want to closer align our codebase to the official svelte ones. That's why we are changing to use single quotes and tabs.

We would still need to switch to not using trailing commas, but we have currently decided against it.


benmccann (approved):
awesome!

AdrianGonz97 (approved):

PR #508, opened 2024-07-25
fix: typescript detection

manuel3108:
In https://github.com/sveltejs/kit/pull/12453 `tslib` got removed, which broke our typescript detection. If we now detect a `tsconfig.json` or a `vite.config.json` in the current or any of the upwards directories, we assume that we have typescript project.

### DIFF
--- a/packages/core/utils/workspace.ts
+++ b/packages/core/utils/workspace.ts
@@ -1,8 +1,10 @@
 import { type AstTypes, parseScript } from '@svelte-add/ast-tooling';
 import { getPackageJson } from './common.js';
-import { commonFilePaths, readFile } from '../files/utils.js';
+import { commonFilePaths, fileExists, findUp, readFile } from '../files/utils.js';
 import { getJsAstEditor } from '@svelte-add/ast-manipulation';
 import type { OptionDefinition, OptionValues, Question } from '../adder/options.js';
+import { remoteControl } from '../internal.js';
+import path from 'path';
 
 export type PrettierData = {
 	installed: boolean;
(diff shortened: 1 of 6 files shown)


packages/core/utils/workspace.ts
```diff
@@ -76,7 +78,22 @@ export async function populateWorkspaceDetails(
...
+			// while executing tests, we only look into the direct `workingDirectory`
+			// as we might detect the monorepo `tsconfig.json` otherwise.
+			tsConfigExists = await fileExists(path.join(workingDirectory, tsConfigFileName));
+			viteConfigExists = await fileExists(path.join(workingDirectory, viteConfigFileName));
+		} else {
+			tsConfigExists = await findUp(workingDirectory, tsConfigFileName);
+			viteConfigExists = await findUp(workingDirectory, viteConfigFileName);
+		}
+
+		workspace.typescript.installed = tsConfigExists || viteConfigExists;
```
AdrianGonz97, line 96: it might be overkill to also be traversing parent directories for the vite config as well. it should be fine to just check the cwd for the `vite.config.ts` and only traverse up for `tsconfig.json`

```suggestion
		let usesTypescript = false;

		if (remoteControl.isRemoteControlled()) {
			// while executing tests, we only look into the direct `workingDirectory`
			// as we might detect the monorepo `tsconfig.json` otherwise.
			usesTypescript = 
				(await fileExists(path.join(workingDirectory, viteConfigFileName))) ||
				(await fileExists(path.join(workingDirectory, tsConfigFileName)));
		} else {
			usesTypescript = 
				(await fileExists(workingDirectory, viteConfigFileName)) ||
				(await findUp(workingDirectory, tsConfigFileName));
		}
		
		workspace.typescript.installed = usesTypescript;
```

packages/core/utils/workspace.ts
```diff
@@ -76,7 +78,19 @@ export async function populateWorkspaceDetails(
...
+
+		if (remoteControl.isRemoteControlled()) {
+			// while executing tests, we only look into the direct `workingDirectory`
+			// as we might detect the monorepo `tsconfig.json` otherwise.
+			usesTypescript ||= await fileExists(path.join(workingDirectory, tsConfigFileName));
+		} else {
+			usesTypescript ||= await findUp(workingDirectory, tsConfigFileName);
+		}
+
+		workspace.typescript.installed = usesTypescript;
```
AdrianGonz97, line 93: taking another look at this, we should probably move this out of the `if(packageJson.devDependencies)` check since it doesn't depend on it anymore

manuel3108: Good catch

AdrianGonz97 (approved):

PR #509, opened 2024-07-26
fix: self closing tags

### LINKED ISSUE #507: End tags added for self-closing elements
### Topic

tailwindcss

### Version

2.3.10

### Expected behavior

`+layout.svelte` shouldn't be touched

### Actual behavior

Screenshot from 2024-07-24 15-29-03

manuel3108:
Closes #507 

We need to enable `xmlMode` of [`dom-serializer`](https://www.npmjs.com/package/dom-serializer) to make this work. According to https://github.com/cheeriojs/dom-serializer/issues/727 otherwise only void html tags like `br` will be self closing.

As you can assume by the naming of `xmlMode` it does not work well if there are nodes containing non xml code like JS or CSS that will be serialized (it produces complete garbage). That's why for svelte files we need to serialize JS, HTML, CSS first and then concatenate them together. 

Using the example given in the issue (`create-svelte` demo app, applying `tailwindcss` adder) and assuming the user opted in to install prettier, this now produces an empty diff (nothing changed). If prettier was not installed I also find the diff more acceptable than previously. 
image

### DIFF
--- a/.changeset/heavy-carpets-roll.md
+++ b/.changeset/heavy-carpets-roll.md
@@ -1,0 +1,5 @@
+---
+'@svelte-add/ast-tooling': patch
+---
+
+fix: self closing tags
--- a/packages/ast-tooling/index.ts
+++ b/packages/ast-tooling/index.ts
@@ -85,7 +85,7 @@ export function parseHtml(content: string) {
 }
 
 export function serializeHtml(ast: Document) {
-	return serializeDom(ast, { encodeEntities: 'utf8', selfClosingTags: true });
+	return serializeDom(ast, { encodeEntities: 'utf8', selfClosingTags: true, xmlMode: true });
 }
 
 export function stripAst<T>(node: T, propToRemove: string): T {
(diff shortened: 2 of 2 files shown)


manuel3108:
Mhmmm, that doesn't work in every case as the tests point out.

image

Then I'm running out of ideas for the moment.

AdrianGonz97:
> Then I'm running out of ideas for the moment.

I wonder if we could adopt https://github.com/xeho91/svelte-ast-print just for the markup portion? we'd strip out the `<script>` and `<style>` tags before giving it to the svelte parser and then hand it off to `svelte-ast-print` for serializing

manuel3108:
I will give that a try!

benmccann:
For this particular case where we've made no changes to the AST, could we avoid re-printing it and writing it to the file?

manuel3108:
Don't know how we could detect that without dramatically uglifying our code.

AdrianGonz97:
the easy (and likely most expensive) way would be to reparse the original content before serialization and do a deep comparison of the two ASTs. then, with something like `magic-string`, we _could_ just replace the content from their respective locations in the text (i.e. if the script is updated, but not the css, then we'd just replace the script text, leaving the css portion alone)

that's just one way of doing it though, I'm not completely sold on it yet

manuel3108:
> I will give that a try!

I don't think this will work. `svelte-ast-print` requires us to use the "modern" `svelte` parser, which seems to be quite different than before. 

I think the only way we can solve this in an acceptable way is by waiting for typescript support in `svelte-ast-print`. I just made an initial attempt and posted it here: https://github.com/xeho91/svelte-ast-print/pull/90

manuel3108:
Not-solvable at this time

PR #518, opened 2024-07-29
feat: Playwright adder

AdrianGonz97:
This was a pretty straightforward adder to implement. The initial implementation is almost entirely pulled from the playwright template from `create-svelte`.

I do have some questions that we should answer before merging:
1. I find the created `tests` directory to be a _bit_ too vague in name. Should we rename it to `e2e`? Alternatively, `integration` could work too
2. Should we also change the npm script to `test:e2e` as well? It's certainly shorter than `test:integration`
3. I'm a bit iffy on the `testDir` and `testMatch` config props. They allow `demo.ts`, `demo.spec.ts`, and `demo.test.ts` to all be valid test files (so long as they reside in the `tests` dir), which feels weird (`demo.ts` is really the odd one out). 
Should we have it only match on `.spec` and `.test` suffixes instead? This would match the default behavior of Playwright, so we'd just have to remove `testMatch` from the config
4. With the introduction of this adder, the spacing between groups has become overwhelming. I've been experimenting with different ways to make them more distinct without spacing. So far, I've landed on bolding the group headers instead:
img
What do you think?


benmccann:
> I find the created tests directory to be a bit too vague in name. Should we rename it to e2e? Alternatively, integration could work too

I don't have a strong opinion either way. I think it doesn't matter too much either way when you only have one type of test there. If you were to have a second kind of test it would depend if you wanted the tests to live alongside each other in the same directory or have a directory per type. I could also see an approach like `tests/e2e`, `tests/integration` if you got to that point

> Should we also change the npm script to test:e2e as well? It's certainly shorter than test:integration

That could make sense. @dummdidumm recently wrote some Svelte 5 testing docs and called them e2e there so it could align with that

> I'm a bit iffy on the testDir and testMatch config props. They allow demo.ts, demo.spec.ts, and demo.test.ts to all be valid test files (so long as they reside in the tests dir), which feels weird (demo.ts is really the odd one out).
Should we have it only match on .spec and .test suffixes instead? This would match the default behavior of Playwright, so we'd just have to remove testMatch from the config

Removing `testMatch` sounds fine to me

> With the introduction of this adder, the spacing between groups has become overwhelming. I've been experimenting with different ways to make them more distinct without spacing. So far, I've landed on bolding the group headers instead:

Looks a little scrunched to me. Honestly, my solution to this would be to remove bootstrap and bulma

AdrianGonz97:
> I think it doesn't matter too much either way when you only have one type of test there. 

yea, this is mainly to cover the confusing case for when you have both playwright and vitest installed

> If you were to have a second kind of test it would depend if you wanted the tests to live alongside each other in the same directory or have a directory per type. I could also see an approach like `tests/e2e`, `tests/integration` if you got to that point

fair point! but at the moment, `create-svelte` (as well as the vitest adder in #519) prescribes that e2e tests go into the `tests` directory, while unit tests with Vitest reside in the `src` directory. if users want to change the locations such that they both could live in `tests` in their own subdirectories, then they certainly could on their own.

In its current form though, we should probably just rename it to `e2e` so it's clearer that "this is the place where playwright tests need to go". I know that I was personally confused by this the first time I ever added both testing frameworks and wondered: "where the heck are my vitest tests?" and "why aren't all of my tests in `tests`?".

> Looks a little scrunched to me. Honestly, my solution to this would be to remove bootstrap and bulma

I think this would just be delaying the inevitable for just a tiny bit longer 😬

benmccann:
> I know that I was personally confused by this the first time I ever added both testing frameworks and wondered: "where the heck are my vitest tests?" and "why aren't all of my tests in tests?".

Ah, that's a good point. I see how that could be a bit confusing. Alright, you've convinced me :smile: 

> I think this would just be delaying the inevitable for just a tiny bit longer 😬

Well there's always some point at which you can overflow. Getting rid of new lines also delays it for only so long as well. I still think it looks nicer with the extra spacing. If we get to a point where it's too untenable to keep adding to the main dialog I think we could potentially split it into multiple ([example](https://github.com/bombshell-dev/clack/issues/197)).

adders/playwright/playwright.svg
manuel3108, line 1: Uhhh, that's an ugly one :D. But that really is their logo, so there we go...

manuel3108 (dismissed):

manuel3108:
Some further notes after testing:
* Should we really add two scripts with the same content? I see that this makes sense if I would also have applied vitest, but I have not done that here. 
```
"test:e2e": "playwright test",
"test": "playwright test",
```
* Should we add the `test-results` folder to the `.gitignore`?
* Personally I would also have expected the test results to be in a folder `.test-results` (prefixed with dot), but I guess that's just personal preference and we are currently using the playwright default. So it's probably fine.


Any reason this is still a draft?

AdrianGonz97:
> * Should we really add two scripts with the same content? I see that this makes sense if I would also have applied vitest, but I have not done that here.

oops, that's supposed to be this:
```
"test:e2e": "playwright test",
"test": "npm run test:e2e",
```
will tweak it

> * Should we add the `test-results` folder to the `.gitignore`?

an oversight on my part, didn't notice that's generated. yea, it should be added to `.gitignore`

> Any reason this is still a draft?

it was a draft so that the questions could be addressed first before merging, but I just forgot to ready it up after :P

manuel3108 (approved):

PR #519, opened 2024-07-30
feat: Vitest adder

AdrianGonz97:
Similar to the Playwright adder in #518, but was a little more involved in the config space to implement. 

This PR is ready to go.


adders/vitest/config/adder.ts
```diff
@@ -0,0 +1,108 @@
...
+					const defineConfigAlias = importSpecifier?.local?.name ?? 'defineConfig';
+					imports.addNamed(ast, 'vitest/config', { defineConfig: defineConfigAlias });
+
+					object.properties(defaultExport.value.arguments[0], { test });
+				} else if (defaultExport.value.type === 'ObjectExpression') {
+					// if the config is just an object expression, just add the property
+					object.properties(defaultExport.value, { test });
+				} else {
+					// unexpected config shape
+					log.warn('Unexpected vite config for vitest adder. Skipping initialization.');
```
benmccann, line 103: "Skipping initialization" makes it sound like the whole adder didn't run, but I think it'd just be this one file
```suggestion
					log.warn('Unexpected vite config for vitest adder. Could not update.');
```

benmccann (approved):
wow!! looks great!!

manuel3108 (dismissed):
Works as expected. 

Same question as in #518, do we need to add the same script twice? Let's discuss this in #518  as this has the same reasoning for sure.

manuel3108 (dismissed):

PR #521, opened 2024-07-31
chore: move `mdsvex` to a different category

manuel3108:


packages/config/adders/official.ts
```diff
@@ -4,8 +4,7 @@ export const adderCategories: AdderCategories = {
 	codeQuality: ['prettier', 'eslint'],
 	css: ['tailwindcss', 'bulma', 'bootstrap'],
 	db: ['drizzle'],
-	markdown: ['mdsvex'],
-	tools: ['storybook', 'routify'],
+	tools: ['storybook', 'routify', 'mdsvex'],
```
AdrianGonz97, line 7: ordering in terms of popularity
```suggestion
	tools: ['storybook', 'mdsvex', 'routify'],
```

.changeset/seven-roses-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@svelte-add/config': patch
+---
+
+chore: move `mdsvex` to a different category
```
AdrianGonz97, line 5: will need to also add `svelte-add` here too

AdrianGonz97 (dismissed):

.changeset/seven-roses-impress.md
benmccann: could we make one a peer dep of the other or something to automatically trigger adding svelte-add when config is updated?

AdrianGonz97: how about fixing them together via changesets instead: https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md

this seems like it's what we're looking for

benmccann:
```
	tools: {
		id: 'tools',
		name: 'Tools',
		description: 'Different tools',
	},
```

I'm not sure I'd typically refer to `mdsvex` as a "Tool". I wonder about maybe calling it "Additional functionality" or something like that?

.changeset/seven-roses-impress.md
manuel3108: As per the bot above it's intending to release that package as well based on that singular changeset

AdrianGonz97: now i'm confused lol - why is it now incrementing the versions of dependents when it doesn't do that for `@svelte-add/adders`?

AdrianGonz97: the only difference that I can see is that `@svelte-add/adders` is a private package, but that's about it?

manuel3108: It's because `config` is a dependency, and `adders` are not. `adders` are just dynamically imported, if i recall correctly.

AdrianGonz97: ah, yea you're right. it's cause `@svelte-add/adders` is a devDependency rather than a dependency, so it doesn't increment. very interesting

packages/config/categories.ts
```diff
@@ -25,19 +25,14 @@ export const categories: CategoryDetails = {
 		name: 'CSS',
 		description: 'Can be used to style your components',
 	},
-	tools: {
-		id: 'tools',
-		name: 'Tools',
-		description: 'Different tools',
+	additionalFunctionality: {
+		id: 'additionalFunctionality',
```
benmccann, line 29: we could make the id be just `additional` to keep it shorter

manuel3108: Good suggestion!

PR #525, opened 2024-07-31
chore: upgrade to typescript-eslint v8 final

benmccann:
did I do the changeset correctly by including only the one package?

### DIFF
--- a/.changeset/eleven-ways-fry.md
+++ b/.changeset/eleven-ways-fry.md
@@ -1,0 +1,5 @@
+---
+'@svelte-add/adders': patch
+---
+
+chore: upgrade to typescript-eslint v8 final
--- a/.changeset/eleven-ways-fry.md
+++ b/.changeset/eleven-ways-fry.md
@@ -1,5 +1,6 @@
 ---
 '@svelte-add/adders': patch
+'svelte-add': patch
 ---
 
 chore: upgrade to typescript-eslint v8 final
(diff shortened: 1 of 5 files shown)


.changeset/eleven-ways-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@svelte-add/adders': patch
```
AdrianGonz97, line 2:
```suggestion
'@svelte-add/adders': patch
'svelte-add': patch
```

package.json
```diff
@@ -42,7 +42,7 @@
 		"rollup-plugin-preserve-shebangs": "^0.2.0",
 		"svelte-add": "workspace:*",
 		"typescript": "^5.5.2",
-		"typescript-eslint": "8.0.0-alpha.37"
+		"typescript-eslint": "8.0.0"
```
AdrianGonz97, line 45:
```suggestion
		"typescript-eslint": "^8.0.0"
```

AdrianGonz97 (commented):
almost! `@svelte-add/adders` is the special case where `svelte-add` also needs to be included in the changeset

however, I think we should probably add `@svelte-add/adders` to the changeset ignore list too since its presence is more confusing than helpful. I don't see a point in maintaining a separate changelog for it when we're always going to need to include `svelte-add` with it anyway

AdrianGonz97 (approved):

manuel3108:
> however, I think we should probably add `@svelte-add/adders` to the changeset ignore list too since its presence is more confusing than helpful. I don't see a point in maintaining a separate changelog for it when we're always going to need to include `svelte-add` with it anyway

Yeah, I think that makes sense. Apart from that everything looks good.

PR #528, opened 2024-08-06
chore: use `defineConfig` for playwright adder

AdrianGonz97:
Didn't realize playwright exported their own `defineConfig` helper function for configs. Using this instead will allow the JS and TS versions of the config be identical.


AdrianGonz97:
Tests are failing due to an unrelated storybook update. It appears that our storybook adder has broken with their most recent `@next` update

manuel3108 (approved):
Works as expected

PR #543, opened 2024-08-09
feat: Lucia Adder

AdrianGonz97:
# WIP

- [x] missing a lot of error handling
- [x] need to gracefully handle existing `user` and `session` tables
- [x] implement named import merging (e.g. append new import specifiers to existing named imports if the sources match)
- [ ] figure out why comments are not preserved for the `app.d.ts` modifications
- [x] still need to implement the logic for `hooks.server.ts`
- [x] add a demo option (mainly based off of the SK example from [lucia's examples repo](https://github.com/lucia-auth/examples/tree/main/sveltekit/username-and-password))
	- to avoid collisions with existing user logic, the demo is created under a new route `/demo` 
	- it features a combined login/register page (`/demo/login`) which mirrors the [Named Actions examples](https://kit.svelte.dev/docs/form-actions#named-actions) in the SvelteKit docs, and a logout action
- [x] link `lucia` to `drizzle`, such that when `lucia` is selected, `drizzle` also gets selected
    - a confirmation prompt was also added so that it can be ignored
- [ ] `db:push` for `better-sqlite3` is throwing an error. might be a bug in the newest drizzle-kit/orm version?
- [ ] cleanup and test

### DIFF
--- a/.changeset/grumpy-camels-return.md
+++ b/.changeset/grumpy-camels-return.md
@@ -1,0 +1,5 @@
+---
+'@svelte-add/core': minor
+---
+
+feat: add `dependsOn` adder option for adders that depend on each other
--- a/.changeset/large-geese-search.md
+++ b/.changeset/large-geese-search.md
@@ -1,0 +1,7 @@
+---
+'@svelte-add/config': minor
+'@svelte-add/adders': minor
+'svelte-add': minor
+---
+
+feat: `lucia` adder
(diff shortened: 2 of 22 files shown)


AdrianGonz97:
that `hooks.server.ts` implementation was quite painful to write 😅. this whole PR needs to be cleaned up a bunch once I get it fully working

benmccann:
going to close this in favor of adding it in the new repo

PR #544, opened 2024-08-10
chore: use vitest

manuel3108:
Superseeds: #480 
In preparation for #461 

Command to test adders locally will slightly change with this:
```
pnpm test tailwindcss
```

will now be
```
pnpm test -t tailwindcss
```

You can also use `pnpm test:ui` which will also spin up vitest ui.

In the future this will allow us to use snapshots to test the outputs of the adders. As vitest now uses proper multi-threading this may also help with speeding up the testing. Currently most time is spent preparing the `storybook` tests.

Here a few crucial notes:
* I was hoping we could just use `vitest` in `@svelte-add/testing-library`. But apparently `vitest` only detects tests in the package you run it in, and not in other dependencies. That's why I'm just passing down the `describe` and `test` methods.
* If you filter the adders to test with `-t tailwindcss` it will still apply all adders, but only run the tests for the adder specified. I did not find an api that allows us to get the current testcase filter
* Since the `-t` option is just a regex filter you could in theory also test only the `kit-js` templates with `-t kit-js`
* #480 did more work, but was pretty hard to review. I decided to create multiple small PR to make things easier (tests for svelte@5 / snapshot tests).


manuel3108:
routify tests are also failing on main.

manuel3108:
superseded by internal PR

PR #555, opened 2024-08-30
feat: unplugin-icons adder

Joex3:
_There is no issue for this change_ (I hope this is okay)

This PR adds an adder for [unplugin-icons](https://github.com/unplugin/unplugin-icons).

When adding, the user is prompted to install an optional icons collection:

grafik

The list can be updated by running `node scripts/update-unplugin-icons-collections`.


manuel3108:
Hey & thanks for your PR.

This looks awesome and works as expected, there is just one minor thing I noticed, although this might come from our tooling.
image

Though, after some internal discussions we have decided that the use-case is pretty niche, so that we don't want to convert this into an official adder. But we do think that this would be a great fit as a community adder!

Community adders are kinda possible right now ([see this template](https://github.com/svelte-add/adder-template)), though they are not promoted in the official cli or anything. As you might have seen the activity in this repository has been getting quite low, that's because we are working on a lot of stuff behind the scenes. 

If you are cool with this, I will get back to you once we have made more progress, to become the first community adder. Would that work for you?

adders/unplugin-icons/collections.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * This file is auto generated and can be updated by running:
```
benmccann, line 2: This is a very partial list, so I'm not sure how we'd decide which get listed here vs omitted. E.g. https://www.npmjs.com/package/@iconify-json/openmoji is missing. You can see all the icon sets at https://icon-sets.iconify.design/

Joex3:
Thanks for the feedback 😄 

> If you are cool with this, I will get back to you once we have made more progress, to become the first community adder. Would that work for you?

Sounds good to me!

PR #556, opened 2024-09-02
feat: Supabase adder

mstibbard:
Started working on a Supabase adder

## Svelte-Add Tasks

- [x] Add `packageManager` to the `nextSteps` data object so instructions can be specific.
- [x] Investigate running scripts within an adder (E.g., `npx supabase init` or `pnpm supabase init`). This needs some thought as currently `svelte-add` only installs packages after all adders run. Executing a script within an adder would require earlier capture of the user's package manager and (optional) earlier package installation

## Supabase Adder Tasks

- [x] Add `cli` option to install Supabase CLI for local development
    - [x] Extend this to run the `${packageManager} supabase init` command
       - [x] If above is achievable, subsequently update the `./supabase/config.toml` with SvelteKit default dev URL, email templates etc.
- [x] Add `demo` option to include the example `/private` routes
   - [x] If `cli` option is also selected, include example `notes` migration and form action etc.
- [x] Add `admin` option to implement a Supabase client w/ Service Role Key in `$lib/server/supabase-admin`
- [x] Add `helpers` option to implement some useful Supabase commands within the user's `package.json`
- [x] Add `auth` multi-select option where the user can choose which auth solutions to add incl.
   - [x] Email + password
   - [x] Magic link / OTP
   - [x] OAuth

### DIFF
--- a/adders/supabase/config/adder.ts
+++ b/adders/supabase/config/adder.ts
@@ -27,13 +27,6 @@ Start your project with a Postgres database, Authentication, instant APIs, Edge 
 			condition: ({ options }) => options.cli,
 		},
 	],
-	scripts: [
-		{
-			description: 'Supabase CLI initialization',
-			args: ['supabase', 'init', '--with-intellij-settings false', '--with-vscode-settings false'],
-			condition: ({ options }) => options.cli,
-		},
-	],
 	files: [
 		{
 			name: () => `.env`,
@@ -564,49 +557,8 @@ Start your project with a Postgres database, Authentication, instant APIs, Edge 
 			},
 			condition: ({ options }) => options.demo && options.cli,
 		},
-		{
-			name: () => './supabase/config.toml',
-			contentType: 'text',
-			content: ({ content }) => {
-				content = content.replace('"http://127.0.0.1:3000"', '"http://localhost:5173"');
-				content = content.replace('"https://127.0.0.1:3000"', '"https://localhost:5173/*"');
-				content = content.replace('enable_confirmations = false', 'enable_confirmations = true');
-
-				content = appendContent(
-					content,
-					dedent`
-						\n# Custom email confirmation template
-						[auth.email.template.confirmation]
-						subject = "Confirm Your Signup"
-						content_path = "./supabase/templates/confirmation.html"
-						`,
-				);
-
-				return content;
-			},
-			condition: ({ options }) => options.cli,
-		},
-		{
-			name: () => './supabase/templates/confirmation.html',
-			contentType: 'text',
-			content: () => {
-				return dedent`
-					<html>
-						<body>
-							<h2>Confirm your signup</h2>
-							<p>Follow this link to confirm your user:</p>
-							<p><a
-								href="{{ .SiteURL }}/auth/confirm?token_hash={{ .TokenHash }}&type=email&next={{ .RedirectTo }}"
-								>Confirm your email</a
-							></p>
-						</body>
-					</html>
-					`;
-			},
-			condition: ({ options }) => options.cli,
-		},
 	],
-	nextSteps: ({ options, packageManager }) => {
+	nextSteps: ({ options }) => {
 		const steps = [
 			'Visit the Supabase docs: https://supabase.com/docs',
 			'Update the authGuard server hook function with your protected routes',
@@ -616,9 +568,9 @@ Start your project with a Postgres database, Authentication, instant APIs, Edge 
 			steps.push(
 				dedent`Local development environment:
 
-					1. Initialize the local development environment: ${colors.yellow(`${packageManager} supabase init`)}
-					2. Start the local development services: ${colors.yellow(`${packageManager} supabase start`)}. This may take a while the first time you run it
-					3. Update ${colors.green('./supabase/config.toml')} [auth] section \`site_url\` and \`additional_redirect_urls\` to use port 5173
+					1. Initialize the local development environment (E.g., ${colors.yellow(`npx supabase init`)} or ${colors.yellow(`pnpm supabase init`)})
+					2. Start the local development services (E.g., ${colors.yellow(`npx supabase start`)} or ${colors.yellow(`pnpm supabase start`)})
+					3. Update ${colors.green('./supabase/config.toml')} [auth] section \`site_url\` and \`additional_redirect_urls\` to use port http://localhost:5173
 					4. Depending on your Auth selections, you may need to create local email templates and update ${colors.green('./supabase/config.toml')}
 					`,
 			);
@@ -661,12 +613,12 @@ function generateEnvFileContent({ content, options }: TextFileEditorArgs<typeof 
 
 function addEnvVar(content: string, key: string, value: string) {
 	if (!content.includes(key + '=')) {
-		content = appendContent(content, `${key}=${value}`);
+		content = appendEnvContent(content, `${key}=${value}`);
 	}
 	return content;
 }
 
-function appendContent(existing: string, content: string) {
+function appendEnvContent(existing: string, content: string) {
 	const withNewLine = !existing.length || existing.endsWith('\n') ? existing : existing + '\n';
 	return withNewLine + content + '\n';
 }
--- a/adders/supabase/config/adder.ts
+++ b/adders/supabase/config/adder.ts
@@ -6,7 +6,7 @@ export const adder = defineAdderConfig({
 		id: 'supabase',
 		name: 'Supabase',
 		description: `Supabase is an open source Firebase alternative.
-Start your project with a Postgres database, Authentication, instant APIs, Edge Functions, Realtime subscriptions, Storage, and Vector embeddings.`,
+Start your project with a Postgres database, authentication, instant APIs, edge functions, realtime subscriptions, storage, and vector embeddings.`,
 		environments: { svelte: false, kit: true },
 		website: {
 			logo: './supabase.svg',
--- a/adders/supabase/config/adder.ts
+++ b/adders/supabase/config/adder.ts
@@ -322,6 +322,31 @@ Start your project with a Postgres database, Authentication, instant APIs, Edge 
 			},
 		},
 		{
+			name: ({ kit, typescript }) =>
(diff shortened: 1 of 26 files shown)


adders/supabase/config/options.ts
```diff
@@ -0,0 +1,25 @@
+import { colors, defineAdderOptions } from '@svelte-add/core';
+
+export const options = defineAdderOptions({
+	demo: {
+		question: 'Do you want to include demo routes to show protected routes?',
+		type: 'boolean',
+		default: false,
+	},
+	admin: {
+		question: `Do you want to add an Supabase admin client? ${colors.red('Warning: This client bypasses row level security, only use server side.')}`,
```
benmccann, line 10: Could it be in a `server` file or directory to enforce it rather than relying on a warning?
```suggestion
		question: `Do you want to add a Supabase admin client? ${colors.red('Warning: This client bypasses row level security, only use server side.')}`,
```

packages/config/categories.ts
```diff
@@ -4,7 +4,7 @@ export type CategoryInfo = {
 	description: string;
 };
 
-export type CategoryKeys = 'codeQuality' | 'css' | 'db' | 'testing' | 'additional';
+export type CategoryKeys = 'codeQuality' | 'css' | 'db' | 'testing' | 'additional' | 'baas';
```
benmccann, line 7: I'd put in under DB rather than creating a new category 
```suggestion
export type CategoryKeys = 'codeQuality' | 'css' | 'db' | 'testing' | 'additional';
```

adders/supabase/config/adder.ts
```diff
@@ -0,0 +1,585 @@
+import { defineAdderConfig, dedent, type TextFileEditorArgs, colors } from '@svelte-add/core';
+import { options as availableOptions } from './options';
+
+export const adder = defineAdderConfig({
+	metadata: {
+		id: 'supabase',
+		name: 'Supabase',
+		description: `Supabase is an open source Firebase alternative.
+Start your project with a Postgres database, Authentication, instant APIs, Edge Functions, Realtime subscriptions, Storage, and Vector embeddings.`,
```
benmccann, line 9:
```suggestion
Start your project with a Postgres database, authentication, instant APIs, edge functions, realtime subscriptions, storage, and vector embeddings.`,
```

adders/supabase/config/adder.ts
```diff
@@ -0,0 +1,585 @@
...
+		const steps = [
+			'Visit the Supabase docs: https://supabase.com/docs',
+			'Update the authGuard server hook function with your protected routes',
+		];
+
+		if (options.cli) {
+			steps.push(
+				dedent`Local development environment:
+
+					1. Initialize the local development environment: ${colors.yellow('pnpm supabase init')}
```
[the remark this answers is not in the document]
manuel3108: Doesn't look like we are exporting this as of now, but we could change that for sure.

packages/config/adders/official.ts
```diff
@@ -6,6 +6,7 @@ export const adderCategories: AdderCategories = {
 	css: ['tailwindcss'],
 	db: ['drizzle'],
 	additional: ['storybook', 'mdsvex', 'routify'],
+	baas: ['supabase'],
```
manuel3108, line 9:
```suggestion
	db: ['drizzle', 'supabase'],
	additional: ['storybook', 'mdsvex', 'routify'],
```
in accordance to Ben's comment below

adders/supabase/config/options.ts
mstibbard: The adder creates the admin client in `$lib/server/supabase-admin` so the warning is somewhat redundant. Happy to remove if that is preferred. I included it because there's nothing stopping an unaware user from just moving it out of that directory and inadvertently ruining themself.

adders/supabase/config/adder.ts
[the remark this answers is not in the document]
mstibbard: That was something I was going to look into today -- include the package manager in the nextSteps data object.

I also want to investigate using the `executeCli` function within an adder. Ideally I can run the `npm/yarn/pnpm/bun supabase init` command as part of the adder (if the user selected `--supabase-cli true`). I want to run that command and then modify the resulting`./supabase/config.toml` file. The alternative is to hardcode the config file but then there is ongoing maintenance in this project when supabase makes changes. Keen for suggestions.

mstibbard:
I've added a **flawed** approach to running scripts within an adder -- it only works within an existing project using pnpm (and if you have `supabase` cli already cached)... It was an experiment that I will reverse out soon.

I'm thinking a less-flawed approach may be to capture the package manager earlier in the flow and if any `adder.scripts` exist then install packages within that adder? However that would result in at least 2 runs of e.g., `npm install`. Would that be acceptable? Is there a better approach?

manuel3108:
> I've added a **flawed** approach to running scripts within an adder -- it only works within an existing project using pnpm (and if you have `supabase` cli already cached)... It was an experiment that I will reverse out soon.
> 
> I'm thinking a less-flawed approach may be to capture the package manager earlier in the flow and if any `adder.scripts` exist then install packages within that adder? However that would result in at least 2 runs of e.g., `npm install`. Would that be acceptable? Is there a better approach?

Why would it only work in a already created package manager that uses pnpm? If its the problem to detect the currently running package manager, we might be able to borrow that code from `create-svelte`: https://github.com/sveltejs/kit/blob/05cfb98d0a93e2a2c90b5647bc617f8479d590b7/packages/create-svelte/utils.js#L56-L70

In the worst case, couldn't we ask the user to select a PM without installing the deps directly in the beginning? Although i would prefer an automatic detections as suggested above.

mstibbard:
I've now implemented a different approach. Keen for your thoughts before I proceed any further.

Thank you for pointing me to that snippet from `create-svelte` -- massive help!

Re: my previous comment which upon a re-read was not clear. I don't know enough about package managers to explain why it worked for pnpm, I assume it was due to caching. The experiment was not flawed re: identifying the package manager. It was flawed because it retained the current flow of `svelte-add` which installs packages _after_ all adders run. This meant attempting to run a dependency script before it was installed, yielding:

- `npx supabase init` -> `Need to install the following packages: supabase@1.191.3 Ok to proceed? (y)`
- `yarn supabase init` -> `error Command "supabase" not found.`
- `pnpm supabase init` -> Succeeds
- `bun supabase init` -> `error: Script not found "supabase"`

manuel3108:
> * `npx supabase init` -> `Need to install the following packages: supabase@1.191.3 Ok to proceed? (y)`
> * `yarn supabase init` -> `error Command "supabase" not found.`
> * `pnpm supabase init` -> Succeeds
> * `bun supabase init` -> `error: Script not found "supabase"`

Did you try `yarn dlx` (https://stackoverflow.com/a/68483894) and `bunx` (https://bun.sh/docs/cli/bunx) and the same should work for `pnpx` or `pnpm dlx`. You could alternatively use `COMMANDS[packageManager].execute` to get the appropriate execution script from `package-manager-detector`

---

Now let's try running it (keep in mind I have never used supabase before)
image
Some of the next steps are indented to much. You might want to use the `dedent` utility that's already used in multiple places.

Looks like the condition for `demo` is not working. At least it did not prompt me for it, until i removed the condition.

On the home page a login page (/auth) is linked, that doesn't seem to exist, thus showing 404. Looks like it has the same condition problem as above. After removing that, i was able to proceed further.

Not sure about the notes on the private page. Are they supposed to be visible anywhere once I send them?

---

As an FYI, we are currently working on a lot of stuff behind the scenes. We don't expect the adder configuration to change in any noticeable amount, so it should be straightforward to prepare this for a merge once we get there. We will let you know for sure!

mstibbard:
Thanks! Swapped to use execute commands, much better. 

I cannot reproduce the `demo` not working though? The demo requires basic auth or magic link because are self-contained. There are so many extra steps with an OAuth provider that it didn't seem realistic for a 'quick demo'. If no auth method is chosen OR only OAuth is chosen the `demo` prompt should not show.

The home page should be untouched unless you pick `demo` (and hence have also picked `basic` or `magicLink` auth). If no auth options are selected then no `/auth` routes are created so it should 404. Did you not select any auth options and then choose demo (as you removed its conditional)?

Re: notes. They should appear as an unordered list directly above the input text box. Is that not working?

manuel3108:
Just rechecked. Bug was sitting in front of the computer. Didn't realize the auth question was a multiselect. Now everything that happened makes sense. 

But does that make sense? How would you structure the demo app with multiple logins? Does that even work.

FYI, will be off for a few days.

mstibbard:
It would be uncommon to choose both basic and magic link but nothing technically stopping it. The magic link flow is the email confirmation flow every time the user logs in. Choosing basic/magic link + OAuth would be pretty common

manuel3108:
Hey @mstibbard sorry for the long wait. Thanks for this awesome PR which was a really good starting point for the supabase adder implementation.

As I mentioned before, we are currently working on some stuff behind the scenes, which means we are working in a private repo 😆. But it's not going to be private for much longer, that's why i wanted to give you a heads up.

Here is what we changed:
* adapted everything to work with the updated codebase
* removed OAuth login
  * it felt strange to only support google OAuth will supabase supports many providers
  * the setup didn't really change that much, and left the user with a long to do list on what he needs to do to make this work. We figured we'd leave that to the supabase docs, they'll for sure do a better job at this
* Made the adder more robust (WIP)
  * The adder, as discussed here, did completely override i.e. `+layout.svelte`, although it was only necessary to add a method and some html snippets. 
  * There were other common files where this happened, but im sure you get the point. As for the `/auth` and `/private` files we left them as is, as they are not super common to have in an existing project

We did all of this by patching in your changes, so GH should credit you once this is merged:
image

We hope that you are cool with all those changes, if not please let me know!

manuel3108:
@mstibbard We had another thought about my last comment, are you using discord by any chance, so that we discuss a few things in private? Would you mind hitting me up? (manuel3108, you can also find me on the svelte discord)

mstibbard:
Sure, I have sent a friend request on discord. I'm in Sydney, Australia so unsure how well timezones will work

PR #557, opened 2024-09-08
feat: add tailwindcss forms option

mstibbard:
Added `@tailwindcss/forms` plugin option on the `tailwindcss` adder + test

Updated the tailwind package versions


benmccann:
Thank you! We'll probably change the question to be a multiselect like "which tailwind plugins would you like to install?". We're working on some updates behind the scenes and will take a look at including this, so will have to close the PR for now. I appreciate the effort though!

